PR #54, opened 2020-04-19
Use radix as redis client

ostcar:
Also switches the field-values of the redis stream.


ostcar:
See https://github.com/OpenSlides/openslides-datastore-service/issues/24

PR #56, opened 2020-05-01
Datastore

ostcar:
Make the autoupdate-service compatible with the datastore


ostcar:
Before I merge this, the race condition in the datastore service has to be fixed. See https://github.com/OpenSlides/openslides-autoupdate-service/blob/a5cdecf0aa9427f329fffcd8c25c0ab666e75f09/internal/datastore/datastore.go#L61-L62

PR #66, opened 2020-06-10
(WIP) Changes required for integrating all services.

FinnStutzenstein:
- Make a normal startup the default case
- changed some environment variables to match in all services
- Added a development docker environment with a hot-reload daemon

@ostcar The changes might not be pretty, but this was not the goal. I am just trying to get everything working together :D

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -22,10 +22,10 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Build autoupdate image
-      run: docker build . --tag autoupdate
+      run: docker build . --tag openslides-autoupdate
 
     - name: Start autoupdate
-      run: docker run -d --network host --env MESSAGING_SERVICE=redis autoupdate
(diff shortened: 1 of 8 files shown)


FinnStutzenstein:
I suspect, since I changed some defaults, the tests must be changed

ostcar:
@FinnStutzenstein All done. Can I merge this?

ostcar:
@FinnStutzenstein I added the environment variable `DATASTORE_READER_PROTOCOL`. The default is `http`. The variable is needed, if we want to create secure connections with https.

Do you want to use it? An alternative is, that `DATASTORE_READER_HOST` is not `localhost` but `http://localhost`. What do you think?

FinnStutzenstein:
I like the extra variable `DATASTORE_READER_PROTOCOL` with the default to http. At least the overlay networks in docker stack are encrypted, so for our deplayment we do not need https, but if someone else is interested, he/she can use the variable.

I did not test the changes, but I think we need another changes regarding keep-alive (see my email for os3-external-autoupdate, this applies here too). -> You can merge this.

PR #102, opened 2020-09-21
restricter: switch to models.yml

ostcar:


ostcar:
After this is merged: https://github.com/OpenSlides/OpenSlides/pull/5572

Correct the models.yml url and merge

PR #132, opened 2020-12-09
New Auth Format

ostcar:


ostcar:
@FinnStutzenstein I merge this so it is easier for you to test it. With this PR the autoupdate-service should work together with the auth-service. Please tell me if it works.

PR #169, opened 2021-03-18
Projector

ostcar:
This adds the projector url to the autoupdate service.

The output format is the one description [here ](https://github.com/OpenSlides/OpenSlides/wiki/Projector-System).

The slide-functions are missing. Data is always `"abc"`.


ostcar:
@FinnStutzenstein I would like to merge this. If you like, you can test or review it. This PR onlyy adds the projector-feature. Everything else should not be changed.

I only created the `user`-Slide, `list_of_speakers`-Slide and `current_list_of_speaker`slide. For everything else see:
* #174 
* #175 
* #176 
* #177 
* #178 
* #179 
* #180 
* #181 
* #182 
* #183 
* #184 
* #185 
* #186 

Maybe this a good issues for new developers to come in touch with go and the autoupdate-service. If you like, I can give an indroduction how to write a slide.

FinnStutzenstein:
Looks good. I only tested the user slide in the client so far, but it seems, that this concept works very well.

One question/hint: What do you do, if something goes wrong while calculating a slide? I expect (as in OS3), that the slide data is `{"error": "<the error>"}`. Does this hold for your implementation? If so, please change all `return []byte(`"TODO"`), nil, nil` to either raise an error (so you make directly usage of this feature), or return this object.

PR #187, opened 2021-04-01
Loading auth secrets from secret file

ostcar:
Reading secets from file and fail, if it does not exist, expect in development mode.


ostcar:
@FinnStutzenstein This is finally ready. The Readme and the readme-test show, that it works as expected.

FinnStutzenstein (approved):

PR #200, opened 2021-04-30
Projector.service user slide

r-peschke:
Perhaps we should take 2 strategies to resolve the template fields:
- the implemented one reads all replacements
- in most cases there is only one replacement that we know, in most cases the meeting_id". We could use this as replacement without extra request to database and without appending the $-field to the list of keys.


internal/projector/slide/user.go
```diff
@@ -35,20 +36,24 @@ func (u dbUser) String(meetingID int) string {
 
 	if level := u.Level[meetingID]; level != "" {
 		parts = append(parts, fmt.Sprintf("(%s)", level))
+	} else if level = u.DefaultLevel; level != "" {
+		parts = append(parts, fmt.Sprintf("(%s)", level))
```
ostcar, line 40: I would do it like this:
```go
level := u.Level[meetingID]
if level == "" {
  level = u.DefaultLevel
}
if level != "" {
  parts = append(parts, fmt.Sprintf("(%s)", level)
}
```

internal/projector/slide/user_test.go
```diff
@@ -10,83 +10,118 @@ import (
...
+			"Title Firstname Lastname Username DefaultLevel",
+			map[string]string{
+				"user/1/username":                `"jonny123"`,
+				"user/1/title":                   `"Dr."`,
+				"user/1/first_name":              `"Jonny"`,
+				"user/1/last_name":               `"Bo"`,
+				"user/1/default_structure_level": `"Switzerland"`,
+			},
+			`{"user":"Dr. Jonny Bo (Switzerland)"}`,
+			emptySlice,
```
ostcar, line 124: This is a bit strange. It is called `emptySlice` but it contains three elements. Maybe you should rename it

internal/projector/slide/user.go
```diff
@@ -35,20 +36,24 @@ func (u dbUser) String(meetingID int) string {
...
 }
 
+func getUserRepresentation(ctx context.Context, ds projector.Datastore, p7on *projector.Projection) (encoded []byte, keys []string, err error) {
+	var u dbUser
+	keys, err = datastore.Object(ctx, ds, p7on.ContentObjectID, &u)
+	if err != nil {
+		return nil, nil, fmt.Errorf("getting user object: %w", err)
+	}
+
+	return []byte(fmt.Sprintf(`{"user":"%s"}`, u.String(p7on.MeetingID))), keys, nil
```
ostcar, line 53: Maybe we have to rename `u.String()` to something else. When I read this, I thought the function would convert the MeetingID to a string.

internal/projector/slide/user_test.go
```diff
@@ -107,9 +143,57 @@ func TestUser(t *testing.T) {
...
+	p7on := &projector.Projection{
+		ContentObjectID: "user/1",
+		MeetingID:       222,
+	}
+
+	bs, keys, err := userSlide.Slide(context.Background(), ds, p7on)
+	assert.Nil(t, bs)
+	assert.Nil(t, keys)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Error encoding responceData `map[user:map[1:map[username:jonny123]]]`")
```
ostcar, line 198: I don't understand why this should be an error. I thought this would return `{"user":"jonny123"}` Can you explain this to me.

ostcar (changes requested):

internal/projector/slide/user.go
r-peschke: Done

internal/projector/slide/user_test.go
r-peschke: Done. It's now addKeysExpected, equal to the related structure field.

internal/projector/slide/user.go
r-peschke: The name `String` was yours. I thought it's a kind of standard in Go to get a string representation of an instance.
Isn't there any standard for this? Do you have a better proposal that we could use as a kind of generic name?

internal/projector/slide/user_test.go
r-peschke: The reason why it's raising the error are the missing  around the "jonny123"-data. I just wanted to produce an error for testing purposes.
Choose another more serious error: based on new specification in [wiki](https://github.com/OpenSlides/OpenSlides/wiki/Projector-System#user) if there isn't minimum one of username, firstName, lastName there is no output and we should raise an error.

internal/projector/slide/user.go
```diff
@@ -28,29 +29,45 @@ func (u dbUser) String(meetingID int) string {
...
 		return full
-	}(u.Title, u.FirstName, u.LastName)
+	}(u.FirstName, u.LastName)
 
+	fmt.Println("user.go L. 34: Username:" + u.Username)
 	if len(parts) == 0 {
-		return u.Username
+		if u.Username != "" {
+			parts = append(parts, u.Username)
+		} else {
```
ostcar, line 38: I don't think this is necessary. Each user has a Username. If not, the database is corrupted.

internal/projector/slide/user.go
```diff
@@ -28,29 +29,45 @@ func (u dbUser) String(meetingID int) string {
...
+	}
+	if level != "" {
 		parts = append(parts, fmt.Sprintf("(%s)", level))
 	}
 
-	return strings.Join(parts, " ")
+	return strings.Join(parts, " "), nil
 }
 
+// Get Representation of the ContentObjectId and MeetingID from the Projection, assuming it's a User
```
ostcar, line 57: Each doc-string in go should start with the name of the function: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

For exported objects this is checked by the linter. But I think it should also be done for non exported objects.

internal/projector/slide/user.go
```diff
@@ -18,7 +18,8 @@ type dbUser struct {
 	DefaultLevel string         `json:"default_structure_level"`
 }
 
-func (u dbUser) String(meetingID int) string {
+// Get instance representation of the user
```
ostcar, line 21: The comment should start with `String `: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

internal/projector/slide/list_of_speakers.go
```diff
@@ -61,9 +61,11 @@ func renderListOfSpeakers(ctx context.Context, ds projector.Datastore, losFQID s
 
 		var user dbUser
 		fetch.Object(ctx, &user, "user/%d", speaker.UserID)
+		var userRepr string
+		userRepr, err = user.String(meetingID)
```
ostcar, line 65: You have to check for the error. And there is no need to define the variable userRepr. Please use:
```go
userRepr, err := user.String(meetingID)
if err != nil {
  return fmt.Errorf("building user representation: %w", err)
}
```

But as written in another comment, I would propose that user.String() does not return an error. The only case is, if the username does not exist and this is not possible.

internal/projector/slide/user.go
```diff
@@ -18,7 +18,8 @@ type dbUser struct {
...
+func (u dbUser) String(meetingID int) (string, error) {
```
ostcar, line 22: The method `String() string` is special. It is defined in the standard package [fmt](https://golang.org/pkg/fmt/#Stringer). The method is called by `fmt.Print`. See: https://play.golang.org/p/XcVK-6EoY7S

But this only works if the signature of String has no arguments and only returns a string.

I know it was me who named the method and gave it the argument. But now I see, that it should be renamed. An alternative would be to find a way to remove the argument and the error-return value.

Do you have an idea for a good name or for a way to use the method without a meetingID?

internal/projector/slide/user.go
```diff
@@ -28,29 +29,45 @@ func (u dbUser) String(meetingID int) string {
...
-	}(u.Title, u.FirstName, u.LastName)
+	}(u.FirstName, u.LastName)
 
+	fmt.Println("user.go L. 34: Username:" + u.Username)
 	if len(parts) == 0 {
-		return u.Username
+		if u.Username != "" {
+			parts = append(parts, u.Username)
+		} else {
+			fmt.Println("user.go L. 39 Errorreturn")
```
ostcar, line 39: Please remove tihs debugging line

internal/projector/slide/user.go
```diff
@@ -28,29 +29,45 @@ func (u dbUser) String(meetingID int) string {
 			full = append(full, s)
 		}
 		return full
-	}(u.Title, u.FirstName, u.LastName)
+	}(u.FirstName, u.LastName)
 
+	fmt.Println("user.go L. 34: Username:" + u.Username)
```
ostcar, line 34: Please remove this debugging line

internal/projector/slide/user.go
```diff
@@ -28,29 +29,45 @@ func (u dbUser) String(meetingID int) string {
...
 func getUserRepresentation(ctx context.Context, ds projector.Datastore, p7on *projector.Projection) (encoded []byte, keys []string, err error) {
 	var u dbUser
 	keys, err = datastore.Object(ctx, ds, p7on.ContentObjectID, &u)
 	if err != nil {
 		return nil, nil, fmt.Errorf("getting user object: %w", err)
 	}
 
-	return []byte(fmt.Sprintf(`{"user":"%s"}`, u.String(p7on.MeetingID))), keys, nil
+	var repr string
+	repr, err = u.String(p7on.MeetingID)
```
ostcar, line 66: You do not need the define the varialbe repr use:

```go
repr, err := u.String(p7on.MeetingID)
```

ostcar (changes requested):

internal/projector/slide/user.go
ostcar: I answered your question in another comment.

internal/projector/slide/user_test.go
ostcar: I would not test the error string. I think the strings are for humans and the test should not fail if the string changes.

If you just want to test that an error happens, then you can check that error is not nil.

If you want to check for a specific error, you should check the error type. `error` is just an interface. Everything with a method `Error() string` can be an error. Most go developers create errors and export them: https://golang.org/src/net/http/server.go?s=64501:64553#L41

I don't line that the error-objects are varialbe (see [this blog post](https://dave.cheney.net/2016/04/07/constant-errors)). And I also don't like to export error objects. This creates unnecessary dependencies between packages. For this reason, I identify errors by methods. See https://github.com/OpenSlides/openslides-vote-service/blob/main/internal/backends/test/test.go#L43-L46

ostcar:
> Perhaps we should take 2 strategies to resolve the template fields:
> 
> * the implemented one reads all replacements
> * in most cases there is only one replacement that we know, in most cases the meeting_id". We could use this as replacement without extra request to database and without appending the $-field to the list of keys.
You are right. I try to solve this in a separate PR. I have an idea...

internal/projector/slide/user_test.go
r-peschke: I'm using now an explicit interface for errors in slides-function. The error will be returned in the expected client format. okay, I expect Finn will agree. What do you think about this way?
The provoked error itself we can remove, because the username since some days is really required. 

About checking the error string I don't agree. We should always check it if there is no own type, constant etc. to identify. Otherwise the test may fail with a new error of the same class without being discovered.

internal/projector/slide/user.go
r-peschke: See above. We should remove it after having a correct error output.

internal/projector/slide/user.go
r-peschke: Done

internal/projector/slide/user.go
r-peschke: Done

internal/projector/slide/list_of_speakers.go
r-peschke: Modified, user.String does not return an error anymore

internal/projector/slide/user.go
r-peschke: Changed the name to StringMeetingDependent.
IMO it is the best, because we need the meeting dependent structure-level for the user representation

internal/projector/slide/user.go
r-peschke: done

internal/projector/slide/user.go
r-peschke: done

internal/projector/slide/user.go
r-peschke: done

FinnStutzenstein:
@ostcar A review regarding the errors would be appreciated

internal/http/http.go
```diff
@@ -114,6 +114,16 @@ func handleError(w http.ResponseWriter, err error, writeStatusCode bool) {
 		return
 	}
 
+	var errSlidesError slidesErrorI
```
ostcar, line 117: I don't like that the http package checks for the slide-error. I think at the border of the projector-app the error should be converted to a `ClientError`

But for me it is ok that you leave the code as it is. I can refactor it later

ostcar (approved):

internal/http/http.go
r-peschke: okay

PR #206, opened 2021-05-10
Projector service for agenda_item_list

r-peschke:
Still not working on reflection error in fetch.go Objects, line  329 with "reflect.Value.Addr of unaddressable value"


ostcar:
I don't like that the function to fetch a list of objects is a separate function. It should be possible to reuse the code. 

I also don't see that this new function is really necessary. You can fall fetch.Objects many times. The only difference is, that you get less database queries. But this should not be a problem, since all keys are in the cache after one user fetched them.

I know that I wrote the "Object" function. But I am not a fan of it. It is too match magic. I don't think we should do more of it.

r-peschke:
Slider for agenda_item_list works as expected with new datastore interface. One test for all possible data for this slide included.

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+	MeetingID       int    `json:"meeting_id"`
+	IsHidden        bool   `json:"is_hidden"`
+	IsInternal      bool   `json:"is_internal"`
+	Depth           int    `json:"level"`
+}
+
+func agendaItemFromMap(in map[string]json.RawMessage) (*dbAgendaItem, error) {
+	bs, err := json.Marshal(in)
+	if err != nil {
+		return nil, fmt.Errorf("encoding agenda item data")
```
ostcar, line 26: Please add the original error:
```
return nil, fmt.Errorf("encoding agenda item data: %w", err)
```

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+
+type dbAgendaItemList struct {
+	AgendaItemIds      []int `json:"agenda_item_ids"`
+	AgendaShowInternal bool  `json:"agenda_show_internal_items_on_projector"`
+}
+
+func agendaItemListFromMap(in map[string]json.RawMessage) (*dbAgendaItemList, error) {
+	bs, err := json.Marshal(in)
+	if err != nil {
+		return nil, fmt.Errorf("encoding agenda item list data")
```
ostcar, line 44: Please add the original error, see above.

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+		fetch := datastore.NewFetcher(ds)
+		defer func() {
+			if err == nil {
+				err = fetch.Error()
+			}
+		}()
+		data := fetch.Object(ctx, []string{"id", "item_number", "content_object_id", "meeting_id", "is_hidden", "is_internal", "level"}, p7on.ContentObjectID)
+		agendaItem, err := agendaItemFromMap(data)
+		if err != nil {
+			return nil, nil, err
```
ostcar, line 66: Please wrap the error so it is easier to find this line in a traceback:

```
  return nil, nil, fmt.Errorf("get agenda item from map: %w", err)
```

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+		}
+		collection := strings.Split(agendaItem.ContentObjectID, "/")[0]
+		titleFunc := store.GetTitleFunc(collection)
+		value := map[string]interface{}{"agenda_item_number": agendaItem.ItemNumber}
+		titleInfo, err := titleFunc(ctx, fetch, agendaItem.ContentObjectID, agendaItem.MeetingID, value)
+		if err != nil {
+			return nil, nil, err
```
ostcar, line 73: Please wrap the error

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+		}
+		responseValue, err := json.Marshal(map[string]interface{}{"title_information": titleInfo, "depth": agendaItem.Depth})
+		if err != nil {
+			return nil, nil, err
```
ostcar, line 77: Please wrap the error

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+		defer func() {
+			if err == nil {
+				err = fetch.Error()
+			}
+		}()
+
+		data := fetch.Object(ctx, []string{"agenda_item_ids", "agenda_show_internal_items_on_projector"}, p7on.ContentObjectID)
+		agendaItemList, err := agendaItemListFromMap(data)
+		if err != nil {
+			return nil, nil, err
```
ostcar, line 97: Please wrap the error

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+		}
+
+		for _, aiID := range agendaItemList.AgendaItemIds {
+			data = fetch.Object(ctx, []string{"id", "item_number", "content_object_id", "meeting_id", "is_hidden", "is_internal", "level"}, fmt.Sprintf("agenda_item/%d", aiID))
```
ostcar, line 101: You don't need `fmt.Sprintf`. Have a look at the signature of the `Object` function: https://github.com/OpenSlides/openslides-autoupdate-service/blob/d00dd3c5df681febcf19b2c7f7051048590e31db/pkg/datastore/fetch.go#L38

You can write `fetch.Object(ctx, []string{...}, "agenda_item/%d", aiID)`

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,128 @@ package slide
...
+			agendaItem, err := agendaItemFromMap(data)
+			if err != nil {
+				return nil, nil, err
```
ostcar, line 104: Please wrap the error. Also on all other places. I don't write it here again but make another review after to added all the error-wrapping.

internal/projector/slide/agenda_test.go
```diff
@@ -0,0 +1,298 @@
...
+		expectKeys []string
+	}{
+		{
+			"Starter",
+			data,
+			`{
+				"items": [
+					{
+						"depth": 0,
+					    "title_information": {
```
ostcar, line 112: The indenting is wrong.

internal/projector/slide/assignment.go
```diff
@@ -2,13 +2,43 @@ package slide
...
 
+type dbAssignment struct {
+	ID    int    `json:"id"`
+	Title string `json:"title"`
+}
+
+func assignmentFromMap(in map[string]json.RawMessage) (*dbAssignment, error) {
+	bs, err := json.Marshal(in)
+	if err != nil {
+		return nil, fmt.Errorf("encoding assignment data")
```
ostcar, line 20: Please wrap the original error.

internal/projector/slide/motion.go
```diff
@@ -2,20 +2,76 @@ package slide
...
+type dbMotion struct {
+	ID     int    `json:"id"`
+	Title  string `json:"title"`
+	Number string `json:"number"`
+}
+
+func motionFromMap(in map[string]json.RawMessage) (*dbMotion, error) {
+	bs, err := json.Marshal(in)
+	if err != nil {
+		return nil, fmt.Errorf("encoding motion data")
```
ostcar, line 21: Please wrap the original error

internal/projector/slide/motion.go
```diff
@@ -2,20 +2,76 @@ package slide
...
+
+type dbMotionBlock struct {
+	ID    int    `json:"id"`
+	Title string `json:"title"`
+}
+
+func motionBlockFromMap(in map[string]json.RawMessage) (*dbMotionBlock, error) {
+	bs, err := json.Marshal(in)
+	if err != nil {
+		return nil, fmt.Errorf("encoding motion data")
```
ostcar, line 39: Please wrap the original error

internal/projector/slide/topic.go
```diff
@@ -2,13 +2,42 @@ package slide
...
 
+type dbTopic struct {
+	ID    int    `json:"id"`
+	Title string `json:"title"`
+}
+
+func topicFromMap(in map[string]json.RawMessage) (*dbTopic, error) {
+	bs, err := json.Marshal(in)
+	if err != nil {
+		return nil, fmt.Errorf("encoding topic data")
```
ostcar, line 20: error wrapping

internal/projector/slide/topic.go
```diff
@@ -2,13 +2,42 @@ package slide
...
+	var m dbTopic
+	if err := json.Unmarshal(bs, &m); err != nil {
+		return nil, fmt.Errorf("decoding topic: %w", err)
+	}
+	return &m, nil
+}
+
 // Topic renders the topic slide.
-func Topic(store *projector.SlideStore) {
+func Topic(store *projector.SlideStore) { // , ds projector.Datastore
```
ostcar, line 31: Please remove the comment.

internal/projector/slides.go
```diff
@@ -39,8 +42,30 @@ type Slider interface {
 
 // SliderFunc is a function that implements the Slider interface.
 type SliderFunc func(ctx context.Context, ds Datastore, p7on *Projection) (encoded []byte, keys []string, err error)
+type TitlerFunc func(ctx context.Context, fetch *datastore.Fetcher, fqid string, meeting_id int, value map[string]interface{}) (title map[string]interface{}, err error)
```
ostcar, line 45: Please change the signature of this function

The return value should not be `map[string]interface{}`. This disables the type system of go. You only use some keys with specific types. Please only return this types. You can either use a struct with the types or return the three keys directly.

internal/projector/slides.go
```diff
@@ -39,8 +42,30 @@ type Slider interface {
...
 
 // Slide calls the func.
 func (f SliderFunc) Slide(ctx context.Context, ds Datastore, p7on *Projection) (encoded []byte, keys []string, err error) {
 	return f(ctx, ds, p7on)
 }
+
+// RegisterTitleFunc registers a function for a collection name.
+func (s *SlideStore) RegisterTitleFunc(collection string, f TitlerFunc) {
```
ostcar, line 53: Please make the names for slides and title more similar. So either rename this to `AddTitleFunc` or rename the old function to `RegisterSlideFunc`

internal/projector/slides.go
```diff
@@ -39,8 +42,30 @@ type Slider interface {
...
+	}
+
+	if _, ok := s.titles[collection]; ok {
+		panic(fmt.Sprintf("GetTitle function for collection %s does already exist", collection))
+	}
+	s.titles[collection] = f
+}
+
+// Get returns a TitleFunc for a name.
+func (s *SlideStore) GetTitleFunc(collection string) TitlerFunc {
```
ostcar, line 65: Please make the name of this function and the body similar to `SlideStore.Get`

pkg/datastore/helper.go
```diff
@@ -0,0 +1,53 @@
+package datastore
+
+import (
+	"fmt"
+	"reflect"
```
ostcar, line 5: Please remove reflect

ostcar (changes requested):

internal/projector/projector.go
```diff
@@ -47,7 +47,7 @@ func Register(ds Datastore, slides *SlideStore) {
 			hotKeys[fqfield] = keys
 			if err != nil {
 				log.Printf("Error parsing slide %s: %v", fqfield, err)
-				bs = []byte(`{"error":"Ups, something went wrong!"}`)
+				bs = []byte(fmt.Sprintf(`{"error":"Error parsing slide %s: %v!"}`, fqfield, err))
```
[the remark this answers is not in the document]
FinnStutzenstein: Errors are not directly shown in the client. If a projection has an error, it is printed to the console and the slide is not shown. Sending the traceback to the client would be appreciated to quickly find errors indicated by the traceback, especially for finding the problem to get around it in live situations. It enormously helps finding small bugs in production to hotfix them.  Additionally, all errors should be printed to the commandline, too. Also there is nothing to hide in these messages: The traceback itself is public. The initial message (e.g. "motion/5 could not be found") should only contain data that would have been published if there was no error (except for relations - but the content is what matters) since every information the code queries is used somehow in the slide data, otherwise you may just not query it.

internal/projector/projector.go
```diff
@@ -86,17 +86,17 @@ func Register(ds Datastore, slides *SlideStore) {
...
 type Projection struct {
-	ID              int    `json:"id"`
-	Type            string `json:"type"`
-	ContentObjectID string `json:"content_object_id"`
-	MeetingID       int    `json:"meeting_id"`
+	ID              int                    `json:"id"`
+	Type            string                 `json:"type"`
+	ContentObjectID string                 `json:"content_object_id"`
+	MeetingID       int                    `json:"meeting_id"`
+	Options         map[string]interface{} `json:"options"`
```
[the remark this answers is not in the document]
r-peschke: I don't understand this. With this change it's really complicate to set the value
agenda_test.go, L. 220 would be: 
`Options:         map[string]json.RawMessage{"only_main_items": []byte("true")},`
and to get agenda.go L. ~110 would be like:
```
			var omi bool
			if err = json.Unmarshal(p7on.Options["only_main_items"], &omi); err != nil {
				return nil, nil, err
			}
```
How could this be written better and in this way, what is it for?

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda.go
r-peschke: Done

internal/projector/slide/agenda_test.go
r-peschke: done

internal/projector/slide/assignment.go
r-peschke: done

internal/projector/slide/motion.go
r-peschke: done

r-peschke: done

internal/projector/slide/motion.go
r-peschke: Done

internal/projector/slide/topic.go
r-peschke: done

internal/projector/slide/topic.go
r-peschke: done

internal/projector/slides.go
r-peschke: What type should it have?
We should give always a json with only the fields for the specified collection back. Otherwise the client has no criteria to distinct, which values to show.

pkg/datastore/helper.go
r-peschke: File removed

internal/projector/slides.go
r-peschke: I don't understand what you mean:
respect to the function name:
Which name do you expect? IMO, if we change something, we should change Get to GetSliderFunc,  because we have 2 kinds of functions to Get.

respect to body:
I think we should check whether the functions is registered for the collection, because it's meant to be extended. AgendaItem/content_object_id always expects a GetTitleInformation-function.

internal/projector/slides.go
r-peschke: Renamed it to RegisterSlideFunc

internal/projector/projector.go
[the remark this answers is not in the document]
ostcar: It does not matter if the error is shown to the client. I don't want, that he can see it in the console or by manually accessing the url. I see this as an security issue and will not accept this change. For the quick fix, you have to look into the server logs.

internal/projector/projector.go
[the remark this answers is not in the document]
ostcar: I had bad experience with `map[string]interface{}`. This is way I use json.RawMessage at all places. I don't want to introduce another style here. 

If you don't want to write the `[]byte` type in test, then you can write converters for example there is the [yaml-converter](https://github.com/OpenSlides/openslides-autoupdate-service/blob/d00dd3c5df681febcf19b2c7f7051048590e31db/pkg/dsmock/datastore_mock.go#L19). I also often use the type `map[string]string` and convert it to `map[string]json.RawMessage`. ([exampe](https://github.com/OpenSlides/openslides-autoupdate-service/blob/d00dd3c5df681febcf19b2c7f7051048590e31db/internal/autoupdate/feature_test.go#L15))

internal/projector/slides.go
ostcar: I though, you would convert it to a string. I understand now, that you return it as a json-object. Maybe the client developers have to say what fields they need.

@FinnStutzenstein It is the function to show the title of an agenda item.

internal/projector/slides.go
ostcar: Yes. This sounds good. Please do this.

internal/projector/slides.go
r-peschke: Not the only, but the most expandable usage of the title information function, is the agenda item list. Each agenda item has a content_object_id and it is it's title, that will be shown. See [here](https://github.com/OpenSlides/OpenSlides/wiki/Projector-System#get_title_informationfqid-meeting_id) the existing collections with getTitleInformation-requirements and their collection-specific content.
I think this is a classical schema to use map[string]interface{} and I would prefer this way:-)

To get it more your way, we could use a pattern with a type struct {type: string, data: json.RawMessage} and could create structs for the individual types, you may see [here](https://www.programmersought.com/article/23385788140/) a small example. Is this an acceptable way to handle this case?

internal/projector/slides.go
r-peschke: Replaced the Get-Function by GetSliderFunc

internal/projector/projector.go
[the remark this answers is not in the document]
r-peschke: I don't think we will create an security issue with this. There is no data involved as @FinnStutzenstein  wrote.

But anywhere, we need a solution. We have to inform the client about the fact that there has been an error, problem or anything else as a reason to not display the content of the projection. What kind of error message we could sent, @ostcar? A special error type to check for or what do you propose?

internal/projector/projector.go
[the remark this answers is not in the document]
r-peschke: I don't wanted to code it with map[string]json.RawMessage, but used a specific type for the options field.

internal/projector/projector.go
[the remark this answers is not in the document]
ostcar: The data is not sent to the user who activated the slide, but to the user who is looking at it. This user can not fix it. If you activate a slide that does not exist, there should be an error message from the backend and not the autoupdate.

internal/projector/projector.go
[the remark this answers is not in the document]
ostcar: Then ping me at the places where the option is used and I code it.

internal/projector/slides.go
ostcar: What I don't like of the interface{} is, that you skip the go-type system. I would prefer if we define specific fields with specific types. I asked @FinnStutzenstein in a mail.

FinnStutzenstein: Generics would be the solution; works fine in the client (typescript) for exactly this type of problem ;)

internal/projector/projector.go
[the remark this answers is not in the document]
r-peschke: I think this is solved with the explicite type ProjectionOptions

internal/projector/projector.go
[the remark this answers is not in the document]
r-peschke: Changed. Instead the slide data, we send now something like  `{"error":"Error parsing slide projection/5/content!"}`

internal/projector/slides.go
r-peschke: I will change the return type of the GetTitleFunction from interface{} to a TitlerFuncResult-struct with specific fields. Marshaling is made with an omitempty, which omits the not used entries depending the type of colletion.

.gitignore
```diff
@@ -3,3 +3,5 @@
 /datastore
 cert
 .vscode
+workspace.code-workspace
+__debug_bin
```
ostcar, line 7: Please add a newline

internal/projector/projector.go
```diff
@@ -47,24 +47,27 @@ func Register(ds Datastore, slides *SlideStore) {
 			hotKeys[fqfield] = keys
 			if err != nil {
 				log.Printf("Error parsing slide %s: %v", fqfield, err)
-				bs = []byte(`{"error":"Ups, something went wrong!"}`)
+				bs = []byte(fmt.Sprintf(`{"error":"Error parsing slide %s!"}`, fqfield))
```
ostcar, line 50: fqfield is not the name of the slide but something like "projection/1/content". I don't see how this message would help, but it also does no harm. If you need it, it is ok for me.

internal/projector/projector.go
```diff
@@ -47,24 +47,27 @@ func Register(ds Datastore, slides *SlideStore) {
...
 				err = nil
 			}
 		}()
 
 		parts := strings.SplitN(fqfield, "/", 3)
 		if len(parts) != 3 {
-			return nil, fmt.Errorf("invalid key %s, expected two '/'", fqfield)
+			err = fmt.Errorf("invalid key %s, expected two '/'", fqfield)
+			return bs, err
```
ostcar, line 58: Please revert that change or explain to me, why you think that it is necessary.

.gitignore
r-peschke: done

internal/projector/projector.go
r-peschke: It is not the name of the slide, but it is the origin data of the requested slide.

internal/projector/projector.go
r-peschke: This is made in conjunction with the defer function, which changes the contents of the variables bs and err. But you are right, the return values are named and we don't need to use the names. I reverted that change

ostcar:
I mad a mess and my github skills are not good enough to fix it. I close this PR here and open a new one. I hope, this will help

PR #209, opened 2021-05-13
Autogenerate models

ostcar:
This uses automatic generated models from the models.yml. It is an alternative to #207


r-peschke (commented):
May be it is a good alternative
Pro: Easy to generate, specialized on model
Missing: Return of keys/tkeys for autoupdate, single given replacements (i.e. meeting_id)

ostcar:
As discussed. We close this for now. If needed, we can add it later.

PR #229, opened 2021-06-17
Projector.service agenda item slide

ostcar:
See #206


internal/projector/projector.go
```diff
@@ -86,17 +97,17 @@ func Register(ds Datastore, slides *SlideStore) {
...
 type Projection struct {
-	ID              int    `json:"id"`
-	Type            string `json:"type"`
-	ContentObjectID string `json:"content_object_id"`
-	MeetingID       int    `json:"meeting_id"`
+	ID              int             `json:"id"`
+	Type            string          `json:"type"`
+	ContentObjectID string          `json:"content_object_id"`
+	MeetingID       int             `json:"meeting_id"`
+	Options         json.RawMessage `json:"options"`
```
r-peschke, line 110: Would be better to have a type than an untyped RawMessage

internal/projector/projector_test.go
```diff
@@ -112,7 +112,24 @@ func TestProjectionUpdateProjectionMetaData(t *testing.T) {
 
 	fields, err := ds.Get(context.Background(), "projection/1/content")
 	require.NoError(t, err, "Get returned unexpected error")
-	expect := `{"id": 0, "content_object_id": "", "type":"projection", "meeting_id": 1}` + "\n"
+	expect := `{"id": 0, "content_object_id": "", "type":"projection", "meeting_id": 1, "options": null}` + "\n"
```
r-peschke, line 115: With options as struct this would result in an option.OnlyMainItems = false, decoding here results in an decode error

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,205 @@ package slide
...
+			p7on.ContentObjectID,
+		)
+
+		agendaItem, err := agendaItemFromMap(data)
+		if err != nil {
+			return nil, nil, fmt.Errorf("get agenda item: %w", err)
+		}
+
+		collection := strings.Split(agendaItem.ContentObjectID, "/")[0]
+		titler := store.GetAgendaTitler(collection)
```
r-peschke, line 90: - AgendaTitler is the wrong name. The GetTitle-method exists for a bunch of collections, agenda is only one of them.
- I think it is better to create the error on one central place, where it is detected. Here you have to code it for all calls to the GetTitle-function and trust in the developer, to check this here.

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,205 @@ package slide
...
+		if titler == nil {
+			return nil, nil, fmt.Errorf("no titler function registered for %s", collection)
+		}
+
+		titleInfo, err := titler.AgendaTitle(ctx, fetch, agendaItem.ContentObjectID, agendaItem.ItemNumber)
```
r-peschke, line 95: The name AgendaTitle is misleading. Using result of the GetTitle-function as AgendaItem-title is only one usage, but there are more independently from AgendaItems. See https://github.com/OpenSlides/OpenSlides/wiki/Projector-System#get_title_informationfqid-meeting_id

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,205 @@ package slide
...
+
+	var ail dbAgendaItemList
+	if err := json.Unmarshal(bs, &ail); err != nil {
+		return nil, fmt.Errorf("decoding agenda item list data: %w", err)
+	}
+	return &ail, nil
+}
+
+type outAgendaItem struct {
+	ContentObjectID  string          `json:"content_object_id"`
```
r-peschke, line 55: This is not the desired output format. All GetTitle-functions have to return the collection-name, see https://github.com/OpenSlides/OpenSlides/wiki/Projector-System#get_title_informationfqid-meeting_id. To identify them, I added the contentObjectId field. IMO it is better to fill the collection name (and the ContentObjectId) on a central place to the returned GetTitle object. Otherwise a developer could forget it, like both of us did in the first approach;-).

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,205 @@ package slide
...
+		)
+		agendaItemList, err := agendaItemListFromMap(data)
+		if err != nil {
+			return nil, nil, fmt.Errorf("get agenda item list: %w", err)
+		}
+
+		var options struct {
+			OnlyMainItems bool `json:"only_main_items"`
+		}
+		if err := json.Unmarshal(p7on.Options, &options); err != nil {
```
r-peschke, line 140: This may produce an error, if there is no typed options-object to unmarshal.

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,205 @@ package slide
...
+			if agendaItem.IsHidden || (agendaItem.IsInternal && !agendaItemList.AgendaShowInternal) {
+				continue
+			}
+
+			if options.OnlyMainItems && agendaItem.Depth > 0 {
+				continue
+			}
+
+			collection := strings.Split(agendaItem.ContentObjectID, "/")[0]
+			titler := store.GetAgendaTitler(collection)
```
r-peschke, line 174: see above, IMO opinion it i sbetter to create this error on one central place

internal/projector/slide/agenda.go
```diff
@@ -2,20 +2,205 @@ package slide
...
+
+			titleInfo, err := titler.AgendaTitle(ctx, fetch, agendaItem.ContentObjectID, agendaItem.ItemNumber)
+			if err != nil {
+				return nil, nil, fmt.Errorf("get title func: %w", err)
+			}
+
+			allAgendaItems = append(
+				allAgendaItems,
+				outAgendaItem{
+					TitleInformation: titleInfo,
```
r-peschke, line 187: see above

internal/projector/slide/agenda_test.go
```diff
@@ -0,0 +1,297 @@
...
+		expect     string
+		expectKeys []string
+	}{
+		{
+			"Starter",
+			data,
+			`{
+				"items": [
+					{
+						"depth": 0,
```
r-peschke, line 111: see desired output format description

internal/projector/slide/agenda_test.go
```diff
@@ -0,0 +1,297 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			closed := make(chan struct{})
+			defer close(closed)
+			ds := dsmock.NewMockDatastore(closed, tt.data)
+
+			p7on := &projector.Projection{
+				ContentObjectID: "meeting/1",
+				Type:            "agenda_item_list",
+				MeetingID:       1,
+				Options:         []byte(`{"only_main_items":true}`),
```
r-peschke, line 219: IMO would be better to use a type than an untyped string

internal/projector/slide/agenda_test.go
```diff
@@ -0,0 +1,297 @@
...
+		expect     string
+		expectKeys []string
+	}{
+		{
+			"Starter",
+			data,
+			`{
+				"depth": 0,
+				"content_object_id": "topic/1",
+				"title_information": {
```
r-peschke, line 262: see desired type description

internal/projector/slide/assignment.go
```diff
@@ -2,13 +2,56 @@ package slide
...
+}
+
 // Assignment renders the assignment slide.
 func Assignment(store *projector.SlideStore) {
-	store.AddFunc("assignment", func(ctx context.Context, ds projector.Datastore, p7on *projector.Projection) (encoded []byte, keys []string, err error) {
+	store.RegisterSliderFunc("assignment", func(ctx context.Context, ds projector.Datastore, p7on *projector.Projection) (encoded []byte, keys []string, err error) {
 		return []byte(`"TODO"`), nil, nil
 	})
+
+	store.RegisterAgendaTitlerFunc("assignment", func(ctx context.Context, fetch *datastore.Fetcher, fqid string, itemNumber string) (json.RawMessage, error) {
```
r-peschke, line 36: misleading name, see above

r-peschke: itemNumber may be the content of this parameter, but this is not necessary

r-peschke: I would prefer to have a typed return value, which could be used and extended (desired collection and content_object_id) on one place.

internal/projector/slides.go
```diff
@@ -2,45 +2,78 @@ package projector
 
 import (
 	"context"
+	"encoding/json"
```
r-peschke, line 5: Please change all the names from Agenda... to something like GetTitle... All these names are used to fulfill the requirement of implementing the GetTitleInformation-function

r-peschke (commented):
I didn't reviewed all files with the GetTitleInformation-function, they are similar to the first one assignment.go

internal/projector/projector_test.go
r-peschke: Fixed in AgendaItemList slider

internal/projector/slide/agenda.go
r-peschke: Fixed

internal/projector/slide/agenda_test.go
r-peschke: fixed

PR #231, opened 2021-06-23
Revert some changes from PR229

r-peschke:


internal/projector/slide/agenda.go
```diff
@@ -137,10 +135,12 @@ func AgendaItemList(store *projector.SlideStore) {
 		var options struct {
 			OnlyMainItems bool `json:"only_main_items"`
 		}
-		if err := json.Unmarshal(p7on.Options, &options); err != nil {
-			return nil, nil, fmt.Errorf("decoding projection options: %w", err)
+		options.OnlyMainItems = false // Default value
```
ostcar, line 138: This is not necessary. Go sets a bool variable to false automatically: https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful

internal/projector/slides.go
```diff
@@ -26,18 +26,18 @@ func (f SliderFunc) Slide(ctx context.Context, ds Datastore, p7on *Projection) (
 	return f(ctx, ds, p7on)
 }
 
-// AgendaTitlerFunc is a function that implements the AgendaTitler interface.
-type AgendaTitlerFunc func(ctx context.Context, fetch *datastore.Fetcher, fqid string, itemNumber string) (json.RawMessage, error)
+// GetTitleInformationType is a type that implements the Titler interface.
+type GetTitleInformationType func(ctx context.Context, fetch *datastore.Fetcher, fqid string, itemNumber string) (json.RawMessage, error)
```
ostcar, line 30: I don't like this name. The a name `FooberFunc` is something used often in go. For example here in the http-lib: https://golang.org/src/net/http/server.go?s=64103:64150#L2055

I would like that this type has the same name as the interface (currently `Titler`) and the sufix `Func`.

ostcar (changes requested):
Please chance the first comment and the second, if you like.

internal/projector/slide/agenda.go
r-peschke: removed it. I wrote it, to set the value explicitly, because I thought about what would be the best default value and the way to set it.

r-peschke:
> Please chance the first comment and the second, if you like.

Deleted the first comment

internal/projector/slides.go
r-peschke: done

ostcar:
Please Ping when you pushed your changes.

PR #232, opened 2021-06-27
Remove debug binary that was introduced by #231

ostcar:


r-peschke (approved):
This debug binary appears regularly on debugging. Perhaps we should put it on the ignore list?

r-peschke (approved):

PR #235, opened 2021-06-29
Sort AgendaItemList by weight (Preordered by backend sort)

r-peschke:


internal/projector/slide/agenda.go
```diff
@@ -168,7 +172,17 @@ func AgendaItemList(store *projector.SlideStore) {
...
+				sortAgendaItems,
+				agendaItem,
+			)
+
+		}
+
+		sort.SliceStable(sortAgendaItems, func(i, j int) bool { return sortAgendaItems[i].Weight < sortAgendaItems[j].Weight })
 
+		var allAgendaItems []outAgendaItem
+		for _, agendaItem := range sortAgendaItems {
```
ostcar, line 185: Is is really necessary to run the for-loop a second time?

I think it should be possible to add the weight field to outAgendaItem and sort the slice at the end of the original for-loop. If you don't want the weight field to be part of the json-output, then you can add the weight field as an unexported field of the outAgendaItem struct.

ostcar (commented):
Only one proposal.

I want to have a response from @FinnStutzenstein before we merge this. I am not sure if he really wants this change.

ostcar:
@FinnStutzenstein Could you test this branch and check, that this is what you want? This sorts all AgendaItems on weight. It does not consider the level.

So if you have a lot of agenda-items on different levels, then you first get all agenda-items with weight 0 and then all agenda items with weight 1 and so on.

For OS3 we have the function [`getFlatTree`](https://github.com/OpenSlides/openslides3-autoupdate-service/blob/0c7fc3c7caf70b974e721f072ba569ce4d762593/internal/apps/agenda/projector.go#L99) that sorts the agenda-tree by weight and level.

r-peschke:
> Only one proposal.
> 
> I want to have a response from @FinnStutzenstein before we merge this. I am not sure if he really wants this change.

It was Finn to demand this sorting and I looked for it.
The backend action [agenda_item.sort](https://github.com/OpenSlides/OpenSlides/wiki/agenda_item.sort) sorts the agenda_item, but as a result of sorting it doesn't store a sorted list or something else, it stores a weight for each each item in a preordered sequence of all items. You can see this best in the [test(https://github.com/OpenSlides/openslides-backend/blob/0515352f7ae230b6ac91b1632894c50b77cc27c8/tests/system/action/agenda_item/test_sort.py#L38).

internal/projector/slide/agenda.go
r-peschke: Very good idea. Done

ostcar:
I see that this code conform with the backend. So it is ok for me, if you merge this.

But I think, that the agenda-item.weight system is broken. See my mail

ostcar (approved):

PR #236, opened 2021-06-29
Projector.service: Topic slider

r-peschke:


internal/projector/slide/topic.go
```diff
@@ -30,7 +32,50 @@ func topicFromMap(in map[string]json.RawMessage) (*dbTopic, error) {
...
+
+		topic, err := topicFromMap(data)
+		if err != nil {
+			return nil, nil, fmt.Errorf("get topic: %w", err)
+		}
+
+		var itemNumber string
+		if topic.AgendaItemID > 0 {
+			itemNumber = fetch.String(ctx, fmt.Sprintf("agenda_item/%d/item_number", topic.AgendaItemID))
+		} else {
```
ostcar, line 61: You can remove the else block.

Go sets the variable "itemNumber" to an empty string. So the else block is a noop

internal/projector/slide/topic.go
```diff
@@ -30,7 +32,50 @@ func topicFromMap(in map[string]json.RawMessage) (*dbTopic, error) {
...
+		)
+
+		topic, err := topicFromMap(data)
+		if err != nil {
+			return nil, nil, fmt.Errorf("get topic: %w", err)
+		}
+
+		var itemNumber string
+		if topic.AgendaItemID > 0 {
+			itemNumber = fetch.String(ctx, fmt.Sprintf("agenda_item/%d/item_number", topic.AgendaItemID))
```
ostcar, line 60: You do not need the fmt.Sprintf here. You can just call:

```go
fetch.String(ctx, "agenda_item/%d/item_number", topic.AgendaItemID)
```

ostcar (changes requested):

internal/projector/slide/topic.go
r-peschke: done

internal/projector/slide/topic.go
r-peschke: done

ostcar (approved):

PR #237, opened 2021-06-29
Can manage users

ostcar:


r-peschke (approved):
Concerning the autoupdate.service OML levels can_manage_organization and can_manage_users are treated identically.

PR #239, opened 2021-06-30
Organization tag

### LINKED ISSUE #227: Cannot get keys for `organization_tags`
```
autoupdate_1        | 2021/06/14 11:10:56 Internal Error: getting keys: get all keys: create keys for keysbuilder: load needed keys: restrict data: checker for key organization/1/organization_tag_ids: check fqids: unknown collection: `organization_tag`
```

If you need more information, please contact me.

ostcar:
Fixes #227

### DIFF
--- a/internal/restrict/permission/collection/simple.go
+++ b/internal/restrict/permission/collection/simple.go
@@ -46,6 +46,26 @@ func isPublic(ctx context.Context, userID int, fqfields []perm.FQField, result m
 	return nil
 }
 
+// LoggedIn can be seen by logged in users.
+func LoggedIn(dp dataprovider.DataProvider, collections ...string) perm.ConnecterFunc {
+	return func(s perm.HandlerStore) {
+		for _, c := range collections {
+			s.RegisterRestricter(c, perm.CollectionFunc(isLoggedIn))
+		}
+	}
+}
+
+func isLoggedIn(ctx context.Context, userID int, fqfields []perm.FQField, result map[string]bool) error {
+	if userID == 0 {
+		return nil
+	}
+
+	for _, field := range fqfields {
+		result[field.String()] = true
+	}
+	return nil
+}
+
 func isInMeeting(dp dataprovider.DataProvider, collection string) perm.CollectionFunc {
 	return func(ctx context.Context, userID int, fqfields []perm.FQField, result map[string]bool) error {
 		return perm.AllFields(fqfields, result, func(fqfield perm.FQField) (bool, error) {
--- a/internal/restrict/permission/tests/testcases/motion/read_stuff.yml
+++ b/internal/restrict/permission/tests/testcases/motion/read_stuff.yml
@@ -12,7 +12,7 @@ fqids:
 - motion_statute_paragraph/1
 
 
-casses:
+cases:
 - name: anonymous
   user_id: 0
   can_see: []
--- a/internal/restrict/permission/tests/testcases/motion/read_stuff.yml
+++ b/internal/restrict/permission/tests/testcases/motion/read_stuff.yml
@@ -15,7 +15,7 @@ fqids:
 casses:
 - name: anonymous
   user_id: 0
-  can_see:
+  can_see: []
 
 - name: with_perm
   permission: motion.can_see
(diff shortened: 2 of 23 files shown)


emanuelschuetze (approved):
tested

internal/restrict/permission/collection/simple.go
```diff
@@ -46,6 +46,26 @@ func isPublic(ctx context.Context, userID int, fqfields []perm.FQField, result m
 	return nil
 }
 
+// LoggedIn can be seen by logged in users.
+func LoggedIn(dp dataprovider.DataProvider, collections ...string) perm.ConnecterFunc {
+	return func(s perm.HandlerStore) {
+		for _, c := range collections {
+			s.RegisterRestricter(c, perm.CollectionFunc(isLoggedIn))
```
r-peschke, line 53: This permits all data for not registered collection-restricters. IMO this is not very secure or is it just a quickfix?

internal/restrict/permission/tests/testcases/motion/read_stuff.yml
```diff
@@ -15,7 +15,7 @@ fqids:
 casses:
```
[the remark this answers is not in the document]
ostcar: Thanks

internal/restrict/permission/collection/simple.go
ostcar: The attribute `collection` is not all collections of OpenSlides, but only the collection the function was called with. See the file `permission/collection.go:22` in this PR.

This function is called once at startup to register the handler for this collections (currently only organization_tag).

The code that does the permission check is in the function `isLoggedIn`at line 58 at this file in this PR. It gets a list of FQFields and returns true for every field the user can see. But this list of FQFields are not all FQFields the user has requested, but only the FQFields that start with the collection, so for example `organization_tag/1/foo` but not `motion/3/foo`.

Does this solve your question? Do you still think this is not secure or a quickfix? If so, please tell me why.

internal/restrict/permission/tests/testcases/motion/read_stuff.yml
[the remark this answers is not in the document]
ostcar: I found the same typo in the projector-permission test. I don't think this yaml tests where a good idea. When I rewrite the permission-package, I will use normal go tests

internal/restrict/permission/collection/simple.go
FinnStutzenstein: > he attribute collection is not all collections of OpenSlides, but only the collection the function was called with.

Can you rename it to avoid name clashes. The term `collection` has a very distinct meaning in OpenSlides so using it for other purposes is a bad idea (see even https://github.com/OpenSlides/OpenSlides/wiki/DE%3AKonzept-OpenSlides-4#19-einheitliche-namensgebung, and `ubiquitous language` from DDD)

r-peschke: @ostcar Your explanation is okay for me and I'll approve the PR. But I'm with @FinnStutzenstein  concerning the term collections.

r-peschke (approved):

internal/restrict/permission/collection/simple.go
ostcar: I don't get it. I I think, that I am using the term `collection` correctly. The Wiki-Page says, that examples for `collections` are `motion` or `item`. So a list of this kind of terms is the plural of `collection`. That is exactly what I am using here.

How would you name a variable that contains a list of string like `motion`, `item` or in this case `organization_tag`?

FinnStutzenstein: Ah ok, sorry. For me it was not clear from your comment. I thought, that there is a different concept there.

r-peschke: I thought you would initiate ALL collections this way and overwrite some of them by registering individually.
Sorry, my fault, I didn't inspect the code in the whole depth.
If you want you could rename collections to something like loginCollections or loginOnlyCollections, but this way it is also okay for me.

r-peschke (approved):

r-peschke:
@ostcar I think you can rebase and merge

PR #246, opened 2021-07-06
Projector.service:slider CurrentSpeakerChyron and error fixes regardi…

r-peschke:
…ng slider choice collection/type


internal/projector/projector.go
```diff
@@ -127,13 +127,16 @@ func (p *Projection) exists() bool {
 	return p.Type != "" || p.ContentObjectID != ""
 }
 
+// slideName extracts the name from Projection.
+// Only collection meeting from ContentObjectID can be used with type
```
ostcar, line 131: Could you add a dot at the end of the docstring? Is it possible, that a word is missing in the sentence? (Or is my english not good enough to understand it?)

internal/projector/slide/list_of_speakers.go
```diff
@@ -156,6 +175,30 @@ func renderListOfSpeakers(ctx context.Context, ds projector.Datastore, losFQID s
 	return b, fetch.Keys(), nil
 }
 
+// get_losID determines the losID and first current_projection of the reference_projector
```
ostcar, line 178: Please change the first word of this docstring to `getLostID`. Please also add a dot at the end.

Maybe the function should be called getLOSID, since LOS is a acronym (https://github.com/golang/go/wiki/CodeReviewComments#initialisms). But maybe we should do an exception here, since the all-caps-word looks strange.

internal/projector/slide/list_of_speakers.go
```diff
@@ -156,6 +175,30 @@ func renderListOfSpeakers(ctx context.Context, ds projector.Datastore, losFQID s
...
+func getLosID(ctx context.Context, ContentObjectID string, fetch *datastore.Fetcher) (losID int, referenceProjectorID int, err error) {
+	parts := strings.Split(ContentObjectID, "/")
+	if len(parts) != 2 || parts[0] != "meeting" {
+		return losID, referenceProjectorID, fmt.Errorf("invalid Content_objectID %s. Expected a meeting-objectID", ContentObjectID)
```
ostcar, line 182: Please either use `conent_object_id` or `ContentObjectID` in the error messages. Also in line 186.

internal/projector/slide/list_of_speakers.go
```diff
@@ -156,6 +175,30 @@ func renderListOfSpeakers(ctx context.Context, ds projector.Datastore, losFQID s
...
+	}
+	referenceProjectorID = fetch.Int(ctx, "meeting/%d/reference_projector_id", meetingID)
+	referenceP7onIDs := fetch.Ints(ctx, "projector/%d/current_projection_ids", referenceProjectorID)
+
+	for _, pID := range referenceP7onIDs {
+		contentObjectID := fetch.String(ctx, "projection/%d/content_object_id", pID)
+		losID = fetch.Int(ctx, "%s/list_of_speakers_id", contentObjectID)
+
+		if losID != 0 {
+			break
```
ostcar, line 196: What happens if all losIDs are 0? The function will return 0. Maybe you should either return an error or document this behavior in the docstring.

internal/projector/slide/list_of_speakers.go
```diff
@@ -197,9 +230,81 @@ func CurrentListOfSpeakers(store *projector.SlideStore) {
 	})
 }
 
+func getCurrentSpeakerData(ctx context.Context, fetch *datastore.Fetcher, losID int, meetingID int) (shortName string, structureLevel string, err error) {
+	data := fetch.Object(ctx, []string{"speaker_ids", "content_object_id", "closed"}, "list_of_speakers/%d", losID)
+	los, err := losFromMap(data)
+	if err != nil {
+		return shortName, structureLevel, fmt.Errorf("loading list of speakers: %w", err)
```
ostcar, line 237: Please return zero values for the first and second return argument, if the last one is an error. The same at the other places in this function.

internal/projector/slide/list_of_speakers.go
```diff
@@ -197,9 +230,81 @@ func CurrentListOfSpeakers(store *projector.SlideStore) {
...
+			if err == nil {
+				err = fetch.Error()
+			}
+		}()
+
+		losID, projectorID, err := getLosID(ctx, p7on.ContentObjectID, fetch)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error in getLosID: %w", err)
+		}
+		meetingID, _ := strconv.Atoi(strings.Split(p7on.ContentObjectID, "/")[1])
```
ostcar, line 279: Maybe chack the error? I know this should always be an int, but maybe it is still a good style to check the error

ostcar (changes requested):

internal/projector/projector.go
r-peschke: done

internal/projector/slide/list_of_speakers.go
r-peschke: done (changed the name in docstring to getLosID and added a dot)

internal/projector/slide/list_of_speakers.go
r-peschke: 0 is a valid return value for losID and it is no error.

internal/projector/slide/list_of_speakers.go
r-peschke: done

internal/projector/slide/list_of_speakers.go
r-peschke: done

internal/projector/slide/list_of_speakers.go
r-peschke: done

ostcar (approved):

Issue #243, opened 2021-06-15
Unlimited reconnection attempts when user in `default`-group


GabrielInTheWorld:
Signing in as a user in the default-group of one meeting (in a meeting) leads to unlimited reconnection attempts to the autoupdate-service.
This has to be investigated.

ostcar:
The error message says, that the field motion_comment_section/1/motion_id does not exist. I also can not find this field in the models.yml or example data

https://github.com/OpenSlides/OpenSlides/blob/610a10712d1194fd56387da07fbc2bac924479b7/docs/example-data.json#L1572

https://github.com/OpenSlides/OpenSlides/blob/610a10712d1194fd56387da07fbc2bac924479b7/docs/models.yml#L1443

Did the models.yml change? Did you inform me? I thought we agreed that a pr that changes the models.yml has to wait until all maintainer changed there services and approved the change.

FinnStutzenstein:
No, this field never existed before since comment sections do not directly link to motions at all.

ostcar:
Ok. I fixed this issue in the autoupdate-service.

But maybe you should also change the client. If there is a 500er on the server and the client will try to reconnect constantly, then it would ddos the server.

Maybe retry 3 times, and once more after a minute but then stop with an error message viable to the user.

PR #248, opened 2021-07-07
Fix issue #243

ostcar:


ostcar:
I fixed this issue without my laptop or editor on the github webpage. It is possible but not recommendable.

internal/restrict/permission/collection/motion.go
```diff
@@ -208,21 +208,12 @@ func (m *motion) canSeeCommentSection(ctx context.Context, userID, id int) (bool
...
+		return false, nil
 	}
 
-	motionOK, err := canSeeMotion(ctx, m.dp, userID, motionID, perms)
-	if err != nil {
-		return false, fmt.Errorf("checking permission for motion: %w", err)
-	}
-	if !motionOK {
-		return false, nil
+	if perms.Has(perm.MotionCanManage) {
```
r-peschke, line 215: IMO it's better to check this before perm.MotionCanSee, because it includes MotionCanSee.

r-peschke (approved):
Approved, but see other comment way down

internal/restrict/permission/collection/motion.go
ostcar: I don't understand. Are you saying it would be faster / better performance?

r-peschke: IMO it's easier to read (and may be also faster). Now I looked for the source of "has" and thought, that it is wrong, because the implementation seems to not respect the permission-tree from permission.yml.
Then I inspected the perm.new and saw that the permission tree is "imported" into the data structure you use to check.
But returning "true" with CanManage-right and checking the CanSee-right only as remaining case, answers this question without knowing the (black-boxed) datastructure with flated permission list.

PR #250, opened 2021-07-08
Send updates on calculcated fields to the client

### LINKED ISSUE #244: No autoupdates for the projector
Sometimes after 10 secs, but e.g. for topics: Changing the topic title does not issue any autoupdates, not even after 10 secs.

ostcar:
Fixes #244

Also update example-data and set a higher cache reset time.


ostcar:
@FinnStutzenstein This should fix the projector-update-issue. But I can not reproduce, that the full autoupdate-service stops working. Maybe you can test it again with this fix and tell me, if there are still problems.

r-peschke (approved):

PR #251, opened 2021-07-09
Fix permission of committee

ostcar:
@FinnStutzenstein could you look at the yaml-file and check, that I correctly interpreted the [wiki page](https://github.com/OpenSlides/OpenSlides/wiki/committee-restrictions)?

### DIFF
--- a/internal/restrict/permission/collection/committee.go
+++ b/internal/restrict/permission/collection/committee.go
@@ -20,33 +20,65 @@ type committee struct {
 	dp dataprovider.DataProvider
 }
 
+// inCommittee returns a set of all committee ids the user is in.
+func inCommittee(ctx context.Context, dp dataprovider.DataProvider, userID int) (map[int]bool, error) {
+	var inCommitteeIDs []int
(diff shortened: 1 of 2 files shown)


FinnStutzenstein (approved):
LGTM

internal/restrict/permission/collection/committee.go
```diff
@@ -20,33 +20,65 @@ type committee struct {
...
+		"description":          true,
+		"meeting_ids":          true,
+		"template_meeting_id":  true,
+		"default_meeting_id":   true,
+		"user_ids":             true,
+		"organization_tag_ids": true,
+		"organization_id":      true,
+	}
+	for _, fqfield := range fqfields {
+		if !committees[fqfield.ID] && orgaLevel != "can_manage_users" {
```
r-peschke, line 73: Assuming the collection of the fqfield is "committee" it seems okay for me.
Don't you want to write a test, where a fqfield isn't accepted, because the user not being part of the committee?

r-peschke (approved):

internal/restrict/permission/collection/committee.go
ostcar: done

PR #252, opened 2021-07-09
Implement projection.poll slider

r-peschke:


FinnStutzenstein:
the key `global_option` is not present, should be always there, even if empty.

r-peschke:
> the key `global_option` is not present, should be always there, even if empty.

From wiki:
`// These keys are only available, if poll/state == "published"
   ...
    global_option
`
In state "published" it is available, even if empty (all values of the struct are set to 0), in the other state it is not existent. Should I change this?

FinnStutzenstein:
Oh, right. My mistake

PR #267, opened 2021-07-18
Restrict user

ostcar:
Changes from https://github.com/OpenSlides/OpenSlides/wiki/user-restrictions

### DIFF
--- a/internal/restrict/permission/collection/user.go
+++ b/internal/restrict/permission/collection/user.go
@@ -30,22 +30,88 @@ func committeeManagerMembers(ctx context.Context, dp dataprovider.DataProvider, 
 		return nil, nil
 	}
 
-	var committeManager []int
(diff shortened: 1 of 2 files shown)


emanuelschuetze (approved):

internal/restrict/permission/collection/user.go
```diff
@@ -327,103 +415,29 @@ var canSeeFields = [...][]string{
...
+	'E': {
+		"committee_ids",
+		"committee_$",
+		"meeting_ids",
+	},
+	'F': {
 		"organization_management_level",
-		"committee_as_member_ids",
-		"committee_as_manager_ids",
 	},
```
tsiegleauq, line 442: @ostcar it seems Group G "password" was missing, or is this handled differently?

https://github.com/OpenSlides/OpenSlides/wiki/user-restrictions

ostcar: That is true. Nobody can see the password. Not even the superadmin. So there was no need to add it.

PR #268, opened 2021-07-19
Projector.service poll change float to string for DecimalFields

r-peschke:


internal/projector/slide/poll.go
```diff
@@ -15,9 +15,9 @@ type optionRepr struct {
 	Text            string          `json:"text"`
 	ContentObjectID string          `json:"content_object_id"`
 	ContentObject   json.RawMessage `json:"content_object"`
-	Yes             *float64        `json:"yes,omitempty"`
-	No              *float64        `json:"no,omitempty"`
-	Abstain         *float64        `json:"abstain,omitempty"`
+	Yes             *string         `json:"yes,omitempty"`     // Python-DecimalField
+	No              *string         `json:"no,omitempty"`      // Python-DecimalField
+	Abstain         *string         `json:"abstain,omitempty"` // Python-DecimalField
```
ostcar, line 20: Are the pointer still necessary? Same question for line 86-88

ostcar (approved):

internal/projector/slide/poll.go
r-peschke: IMO it is more secure and better readable. If there is nothing to read, if there is a string containing  a decimal, I don't "inspect/look into" the decimal field/sting, to see whether it is a correct implementation of whatever decimal number.

PR #269, opened 2021-07-19
Projector.service: Remove agenda_item slide and GetTitleInformation

r-peschke:


ostcar:
@normanjaeckel can this really be removed?

FinnStutzenstein:
Please review, so it can be removed.

ostcar:
Cloud you explain to me how the features that the agenda view had in os3 are implemented in os4 without the agenda view?

Is it still possible to see the agenda? It is still possible to see the title of one item with all its child items?

FinnStutzenstein:
As far as I can see it was already removed in OS3 in the client, but not in the server code. There is no place an `agenda_item` slide is projected, even in the current OS3 client;. You can search for `getBasicProjectorElement`. For the agenda, only the pseudocollection `'agenda/item-list'` is used, which is the list, but not a single item.

emanuelschuetze:
Agenda item slide (not topic slide) was removed from OS2 to OS3.

ostcar:
@emanuelschuetze not from the server side:

OS3-autoupdate: https://github.com/OpenSlides/openslides3-autoupdate-service/blob/aea497e50068f560d54ed4a4b9ceb8695d9ac025/internal/apps/agenda/projector.go#L25

OS3.3: https://github.com/OpenSlides/OpenSlides/blob/a37e2196b3b321a42149e92241f673776b399bfa/server/openslides/agenda/projector.py#L71

Is it still possible in os3 (and will be in os4) so see a list of agenda items?

FinnStutzenstein:
> Is it still possible in os3 (and will be in os4) so see a list of agenda items?

Yes, already implemented with the `agenda_item_list` slide.

> not from the server side:

We have missed it at some point. My guess is some UI-restructure where we removed the slide from the client and forgot about the server. Again, there is no way to use the slide in OS3, so it is dead code, which can be removed.

ostcar:
> > Is it still possible in os3 (and will be in os4) so see a list of agenda items?
> 
> Yes, already implemented with the `agenda_item_list` slide.

Ahh thank you. This was my mistake. I did not realise, that agenda_item_slide and agenda_item_list_slide are two different slides.

ostcar (approved):

PR #270, opened 2021-07-19
Break context when fetching data

ostcar:
The data is not only for one session, so it should not depend on
one context

### DIFF
--- a/internal/projector/projector.go
+++ b/internal/projector/projector.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"log"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


ostcar:
@FinnStutzenstein This PR is not ready. And I don't understand the bug. But it could not reproduce the error anymore after making this change. Please tell me if this fixes the bug for you.

ostcar:
Sry, does not work...

ostcar:
I merge this. This does not 100% solve the error, but the tabs are independent now. If an error happens calculating a projection, the service does not stop working anymore. Please tell me if this solved your issues.

PR #272, opened 2021-07-20
Cache cleanup

ostcar:
This makes the cache code more readable.


r-peschke (approved):
That's to much for my go-expertise. I'll approve it, but we should learn from it, see my signal message

PR #276, opened 2021-07-22
Projector.service Implement motion_block slider

r-peschke:


internal/projector/projector.go
```diff
@@ -132,3 +132,20 @@ func (p *Projection) slideName() (string, error) {
 	}
 	return parts[0], nil
 }
+
+// GetProjection as convenience function reading projections
```
ostcar, line 136: Why do you need this function? I can not find any place (expect the tests) where it is used.

If it is not used, then please remove this function and add it again, when/if you need it

ostcar (changes requested):

ostcar:
Only one comment. The rest is ok and can be merged.

internal/projector/projector.go
r-peschke: removed

PR #279, opened 2021-07-27
New fetcher api

ostcar:
@r-peschke Please have a look at this.

In the first commit, I changed the signature of the fetcher. Now there are two functions to fetch values, one that returns an error if the key does not exist and one, that ignores that error.

I changed the signature of Int, Ints, String to work with both version of the fetch function. See the Examples in the tests.

In the second commit, I changed the signature of the Slider interface in the projector package and started to fix the projector code.

If you are ok with this changes, then I can fix the other slides as well.

### DIFF
--- a/pkg/datastore/fetch.go
+++ b/pkg/datastore/fetch.go
@@ -3,6 +3,7 @@ package datastore
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 )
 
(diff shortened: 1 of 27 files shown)


pkg/datastore/fetch.go
```diff
@@ -34,68 +35,93 @@ func NewFetcher(ds Getter) *Fetcher {
...
+	if f.err != nil {
+		return nil
+	}
+
+	object, keys, err := object(ctx, f.ds, fqID, fields)
+	if err != nil {
+		f.err = fmt.Errorf("fetching object %s: %w", fqID, err)
+		return nil
+	}
+	f.keys = append(f.keys, keys...)
```
ostcar, line 85: This has to be above the error check!

pkg/datastore/fetch.go
```diff
@@ -34,68 +35,93 @@ func NewFetcher(ds Getter) *Fetcher {
...
-// Int fetches an integer from the datastore.
-func (f *Fetcher) Int(ctx context.Context, keyFmt string, a ...interface{}) int {
-	var i int
-	f.Value(ctx, &i, keyFmt, a...)
-	return i
-}
+// FetchIfExist is like Fetch but does not return an error, if a key does not
+// exist. In this case, value is nil.
+func (f *Fetcher) FetchIfExist(ctx context.Context, value interface{}, keyFmt string, a ...interface{}) {
+	if f.err != nil {
```
r-peschke, line 59: You can remove this, it is included in Fetch

r-peschke (approved):

PR #282, opened 2021-07-28
Recognise deadlineExceeded as timeout and increase deadline time for …

r-peschke:
…calculated fields

### DIFF
--- a/pkg/datastore/datastore.go
+++ b/pkg/datastore/datastore.go
@@ -200,7 +200,7 @@ func (d *Datastore) loadKeys(keys []string, set func(string, []byte)) error {
 }
 
 func (d *Datastore) calculateField(field string, key string, updated map[string]json.RawMessage) []byte {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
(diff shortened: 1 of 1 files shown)


pkg/datastore/datastore.go
```diff
@@ -200,15 +200,15 @@ func (d *Datastore) loadKeys(keys []string, set func(string, []byte)) error {
 }
 
 func (d *Datastore) calculateField(field string, key string, updated map[string]json.RawMessage) []byte {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
ostcar, line 203: Why do you need 3 seconds to calculate a field? What is a good time?

r-peschke: If there are slow connections, overloaded server etc., 3 seconds gives a better chance to succeed and if, IMO it's a good "investment". This just disturbes the user in case of an error. 
And usually it doesn't slow down anyway.

r-peschke:
I'll close this, but code to recognize the deadline error transferred to [PullRequest285](https://github.com/OpenSlides/openslides-autoupdate-service/pull/285)

Issue #278, opened 2021-07-26
Slides cannot be projected


FinnStutzenstein:
Every time I project the CLOS, I get the infamous
```
Error: calculating key projection/16/content: fetching projection 16 from datastore: fetching data: getOrSet for keys `[projection/16/id projection/16/type projection/16/content_object_id projection/16/meeting_id projection/16/options]`: waiting for key projection/16/options: waiting for value: context deadline exceeded
```

ostcar:
@FinnStutzenstein could you test again after the fix from @r-peschke in #281 ?

FinnStutzenstein:
This is not fixed. I get a very similar error (`type` instead of the field `options`):
```
Error: calculating key projection/39/content: fetching projection 39 from datastore: fetching object projection/39: fetching data: getOrSet for keys `[projection/39/id projection/39/type projection/39/content_object_id projection/39/meeting_id projection/39/options]`: waiting for key projection/39/type: waiting for value: context deadline exceeded
```
This is the datastore content for the projections:
```
{"id":39,"stable":false,"options":{},"meeting_id":1,"content_object_id":"topic/1","current_projector_id":1}
```

FinnStutzenstein:
Also note, that it does not only appears for the CLOS slide; as you can see this is a topic slide. I also get the error for mediafile slides and more.

r-peschke:
@FinnStutzenstein  I assume that this is the error from log, not the client response.
The error signals a timeout (deadlineExceeded), which is set exclusively for calculated fields to 1 second. 
Debugging the code I can reproduce the error.
With [PR282](https://github.com/OpenSlides/openslides-autoupdate-service/pull/282) I increased the time to 3 seconds and the response contains the timed- out part in the message.

With `git fetch origin pull/282/head:PR282_branch` you could create a branch and test it locally without being merged.

FinnStutzenstein:
I used your branch, but it does not help. There is still an error.

FinnStutzenstein:
To reproduce:
- start the setup in the main repo: `make run-dev`
- open the client
- go the the default committee and create a new meeting
- go into the meeting->agenda. Create a new agenda item
- go to the agenda list and project the item on the default projector
- open the default projector -> error

ostcar:
There was a bug in the topic slide. I fixed it with #284. Afterwards I could not reproduce the error in the way you described.

Could you check if it is still possible to get the bug in other ways?

[...]

PR #284, opened 2021-07-29
fix projector topic. Agenda.item_number is optional

ostcar:
See #278

### DIFF
--- a/internal/projector/slide/topic.go
+++ b/internal/projector/slide/topic.go
@@ -48,7 +48,7 @@ func Topic(store *projector.SlideStore) {
 
 		var itemNumber string
 		if topic.AgendaItemID > 0 {
-			itemNumber = datastore.String(ctx, fetch.Fetch, "agenda_item/%d/item_number", topic.AgendaItemID)
+			itemNumber = datastore.String(ctx, fetch.FetchIfExist, "agenda_item/%d/item_number", topic.AgendaItemID)
 		}
 		if err := fetch.Err(); err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


internal/projector/slide/topic.go
```diff
@@ -48,7 +48,7 @@ func Topic(store *projector.SlideStore) {
 
 		var itemNumber string
 		if topic.AgendaItemID > 0 {
-			itemNumber = datastore.String(ctx, fetch.Fetch, "agenda_item/%d/item_number", topic.AgendaItemID)
+			itemNumber = datastore.String(ctx, fetch.FetchIfExist, "agenda_item/%d/item_number", topic.AgendaItemID)
```
r-peschke, line 51: fetch.Fetch should be sufficient. There should be an error, if there is no object at all for the given AgendaItemID, but not, if the field item_number is empty. There are numerous cases this way.

ostcar: The autoupdate service works on keys/fqfields. There is no difference if an element or a field does not exist. If your code want's to check if an element exists, you have to fetch element/ID/id.

If there are numerous cases where this is wrong, all of them have to be fixed.

If the bug from Finn does still exist, then please tell me how to reproduce it.

ostcar: If you want, we could add a method fetch.FetchIfElemetExist that also fetches the id field and only skips the error of the id does exist but the other field does not.

r-peschke (approved):

r-peschke:
Approved and merged as template for more fields like this

PR #293, opened 2021-08-03
Changes auth-dev-keys

### LINKED ISSUE #291: Change Auth-Dev-Keys
According to OpenSlides/openslides-auth-service#92 change the Auth-Dev-Keys to
- for tokens: `auth-dev-token-key`
- for cookies: `auth-dev-cookie-key`

GabrielInTheWorld:
Fixes #291

### DIFF
--- a/cmd/autoupdate/main.go
+++ b/cmd/autoupdate/main.go
@@ -34,7 +34,8 @@ func main() {
 	}
 }
 
-const debugKey = "auth-dev-key"
+const debugTokenKey = "auth-dev-token-key"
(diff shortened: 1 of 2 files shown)


ostcar:
I also use the keys in the ICC and the vote service. Therefore I need a more generic solution. I will do it today

GabrielInTheWorld:
> I also use the keys in the ICC and the vote service. Therefore I need a more generic solution. I will do it today

Okay.

ostcar:
See #294

PR #294, opened 2021-08-03
Export debug keys

ostcar:


ostcar:
As soon as this is merged, the icc and vote service have to import the new version.

r-peschke (approved):

GabrielInTheWorld:
Looks good!

GabrielInTheWorld (approved):

GabrielInTheWorld:
Please merge and import the new version in the icc- and vote-service and then ping me, that I merge it in the auth-service.

ostcar:
@GabrielInTheWorld https://github.com/OpenSlides/openslides-icc-service/pull/10

The vote service does not need an update since it is currently not used.

Maybe we should move the go-auth-lib into the auth-repo? It is only this directory: https://github.com/OpenSlides/openslides-autoupdate-service/tree/master/pkg/auth

I don't think the auth lib has to be in the autoupdate-repo

GabrielInTheWorld:
> @GabrielInTheWorld [OpenSlides/openslides-icc-service#10](https://github.com/OpenSlides/openslides-icc-service/pull/10)
> 

Nice!

> The vote service does not need an update since it is currently not used.
> 
> Maybe we should move the go-auth-lib into the auth-repo? It is only this directory: https://github.com/OpenSlides/openslides-autoupdate-service/tree/master/pkg/auth
> 
> I don't think the auth lib has to be in the autoupdate-repo

Yes, I think so, too!

Maybe the AU-service and the ICC-service (and later the vote-service, too) could also request the auth-debug-keys from the auth-service. Then we have one central place to manage those keys. It is one request more we have to do, but this is anyway for the development-environment.
What do you think?

PR #298, opened 2021-08-15
cache rewrite

ostcar:


ostcar:
I merge this. This makes it easier to talk about the differences with the version from sascha.

PR #303, opened 2021-08-17
Handle error

ostcar:


internal/projector/slide/motion.go
```diff
@@ -470,8 +470,13 @@ func fillAmendments(ctx context.Context, fetch *datastore.Fetcher, motion *dbMot
 		if err != nil {
 			return fmt.Errorf("motionFromMap: %w", err)
 		}
+
 		fillAmendmentParagraphs(ctx, fetch, motionAmend)
-		fillChangeRecommendations(ctx, fetch, motionAmend)
+
+		if err := fillChangeRecommendations(ctx, fetch, motionAmend); err != nil {
```
r-peschke, line 476: What do you think of a new method of fetcher to add an error to the fetcher. We could write the code more clearly, when methods with fetchers provide errors in fetcher and not on both ways?

ostcar: I don't like the idea. I am still skeptical, that the fetcher-methods do not return an error. I think it is to easy to forget to check the error.

For example: I forgot to check the fetcher error here: https://github.com/OpenSlides/openslides-autoupdate-service/pull/305/files#diff-50930fd2514011aa3568fbd55ebd212c34396e407228168872e7de21f549e902L30

The result was an error message `See returned unexpected error: checking organization management level: getting oml of user 1: meeting/1 does not exist.` This was confusing, because `HasOrganizationManagementLevel`does not read the meeting. But it returned the error message from the fetcher from another place.

I think there are more places where I checked the fetcher error at the wrong place and we will get misleading error messages.

But currently, you only have to think about checking the error when you call fetch.X(). With your idea, you would have to remember to check the error after each function that has fetch as an argument.

PR #322, opened 2021-09-22
Refactor connection filter

### LINKED ISSUE #321: The AU-Service does not send deleted models anymore
I have noticed, that the AU-Service did not send updates when a model was deleted (like `<model/fqid>: null`).
Is there any reason for this behaviour?

ostcar:
fixes #321


r-peschke (approved):

GabrielInTheWorld (approved):
With this PR, the null-values are sent again.

PR #335, opened 2021-10-12
fixed published polls

### LINKED ISSUE #334: Normal user gets no poll result
* Create a new motion poll (e.g. named voting) as admin for delegates and admins
* start voting
* vote with admin user
* vote with delegate user
* stop and publish voting as admin
=> delegate user gets incomplete vote result! Yes/No/Abstain is empty (no autoupdate). Valid votes are available only. 
* open detail view of motion poll: single votes and entitled users are succuessfully visible in both list.

ostcar:
Fixes #334


internal/restrict/collection/option.go
```diff
@@ -36,6 +36,38 @@ func (o Option) see(ctx context.Context, fetch *datastore.Fetcher, mperms *perm.
 	return see, nil
 }
 
+func (o Option) modeB(ctx context.Context, fetch *datastore.Fetcher, mperms *perm.MeetingPermission, optionID int) (bool, error) {
```
GabrielInTheWorld, line 39: What is `modeB`? A better name would start with the prefix `is` or `has`, because this function returns a boolean.

internal/restrict/collection/option.go
```diff
@@ -36,6 +36,38 @@ func (o Option) see(ctx context.Context, fetch *datastore.Fetcher, mperms *perm.
...
+	}
+
+	canManage, err := Poll{}.manage(ctx, fetch, mperms, pollID)
+	if err != nil {
+		return false, fmt.Errorf("checking see poll %d: %w", pollID, err)
+	}
+
+	if canManage {
+		return true, nil
+	}
```
GabrielInTheWorld, line 61: Do you cannot get the permissions of a user just one time?

ostcar: canManage is not a permission. It is the result of the manage-property described here: https://github.com/OpenSlides/OpenSlides/wiki/poll-restrictions#manage-property

The permissions are calculated from the mperm attribute, that is some sort of cache that only calculates the permission once per meeting.

I don't see where I get permissions twice.

PR #337, opened 2021-10-13
Revert change from #322.

ostcar:
It is an expected behaviour, that the client does not gets updates for deleted
objects, if the object was only indirectly requested through a relation-list-field

See #321

This PR also updates some old test code.

### DIFF
--- a/pkg/dsmock/datastore_server.go
+++ b/pkg/dsmock/datastore_server.go
@@ -104,16 +104,8 @@ func (d *DatastoreServer) Update(ctx context.Context) (map[string][]byte, error)
 }
 
 // Send sends keys to the mock that can be received with Update().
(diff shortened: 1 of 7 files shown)


pkg/dsmock/datastore_server.go
```diff
@@ -104,16 +104,8 @@ func (d *DatastoreServer) Update(ctx context.Context) (map[string][]byte, error)
...
-	for k, v := range values {
-		conv[k] = nil
-		if v != "" {
-			conv[k] = []byte(v)
-		}
-
-	}
-	d.c <- conv
+func (d *DatastoreServer) Send(values map[string][]byte) {
+	d.c <- values
```
GabrielInTheWorld, line 108: What is `d.c`?

Issue #356, opened 2021-11-05
Fqid `user/option_ids` is unknown


GabrielInTheWorld:
I get this error
```
docker-autoupdate-1        | 2021/11/05 09:09:09 Internal Error: getting next message: creating first time data: create keys for keysbuilder: load needed keys: restricting data: getting restict func: fqfield "user/option_ids" is unknown, maybe run go generate ./... to fetch all fields from the models.yml
```

This is my request
```js
{
    "collection": "meeting",
    "ids": [
        1
    ],
    "fields": {
        "description": null,
        "name": null,
        "start_time": null,
        "end_time": null,
        "is_active_in_organization_id": null,
        "id": null,
        "poll_ids": {
            "type": "relation-list",
            "collection": "poll",
            "fields": {
                "entitled_group_ids": null,
                "state": null,
                "title": null,
                "type": null,
                "pollmethod": null,
                "content_object_id": null,
                "voted_ids": null,
                "votescast": null,
                "votesinvalid": null,
                "votesvalid": null,
                "option_ids": {
                    "type": "relation-list",
                    "collection": "option",
                    "fields": {
                        "vote_ids": {
                            "type": "relation-list",
                            "collection": "vote",
                            "fields": {
                                "delegated_user_id": null,
                                "option_id": null,
                                "user_id": null,
                                "value": null,
                                "weight": null,
                                "user_token": null,
                                "id": null
                            }
                        },
                        "poll_id": null,
                        "content_object_id": {
                            "type": "generic-relation",
                            "fields": {
                                "title": null,
                                "first_name": null,
                                "last_name": null,
                                "username": null,
                                "vote_weight_$": {
                                    "type": "template"
                                },
                                "structure_level_$": {
                                    "type": "template"
                                },
                                "number_$": {
                                    "type": "template"
                                },
                                "email": null,
                                "gender": null,
                                "is_active": null,
                                "is_physical_person": null,
                                "is_present_in_meeting_ids": null,
                                "last_email_send": null,
                                "default_number": null,
                                "default_structure_level": null,
                                "default_vote_weight": null,
                                "comment_$": {
                                    "type": "template"
                                },
                                "about_me_$": {
                                    "type": "template"
                                },
                                "comment": null,
                                "default_password": null,
                                "id": null
                            }
                        },
                        "yes": null,
                        "no": null,
                        "abstain": null,
                        "id": null
                    }
                },
                "onehundred_percent_base": null,
                "global_option_id": {
[...]

PR #357, opened 2021-11-05
Restrict generic relation fields

ostcar:


GabrielInTheWorld (approved):
This fixes the error in #356 on the client-side.

PR #368, opened 2021-11-21
Switch List of speaker restricter

### LINKED ISSUE #359: Restrictor: List_of_speakers can only be seen if the content_object_id can be seen
Also go though all possible content_object_ids and remove the reference to list_of_speakers

ostcar:
Befor this, you could see the title of an object
if you could see the list_of_speakers

After this, you can only see a list_of_speakers
if you can see the content_object_id

Fixes #359


ostcar:
After https://github.com/OpenSlides/openslides-backend/issues/1038 is fixed, there has to be called `go generate ./...` here.

PR #397, opened 2022-01-09
Trace

ostcar:


jsangmeister:
@ostcar how much work would it be to update this PR? it would be nice to have OTel availble in the autoupdate service, especially for debugging things like the current `poll.stop` performance

ostcar:
I close this PR since it is 1,5 years old and I have to open a new one anyway.

We can continue the discussion here: https://github.com/OpenSlides/openslides-autoupdate-service/issues/401

PR #427, opened 2022-02-23
Update models

ostcar:
See https://github.com/OpenSlides/openslides-backend/issues/1230#event-6111278104


ostcar:
There was something wrong here

PR #431, opened 2022-03-02
Multiple votes accumulated on one person

MJJojo97:
See [https://github.com/OpenSlides/OpenSlides/pull/6376](url)

(After merging, icc-service and vote-service need to link to the corresponding version.)


ostcar:
I updated all autogenerated files in https://github.com/OpenSlides/openslides-autoupdate-service/pull/439

PR #440, opened 2022-03-15
History

ostcar:


internal/restrict/history.go
```diff
@@ -0,0 +1,148 @@
...
+	oldGetter     datastore.Getter
+}
+
+// NewHistory initializes a History object.
+func NewHistory(userID int, current datastore.Getter, old datastore.Getter) History {
+	return History{userID, current, old}
+}
+
+// Get returns the keys the user can see.
+func (h History) Get(ctx context.Context, keys ...string) (map[string][]byte, error) {
```
ostcar, line 30: @normanjaeckel This is the function we talked about. Is this correct? You can find out last comments on the last commit

ostcar:
@emanuelschuetze I changed the permission to see the old data to "is admin". I think this can be merged

PR #450, opened 2022-03-27
Introduce setting allowing to enable or disable max_votes_per_option

MJJojo97:


ostcar:
This PR is not necessary. I have to wait until the models.yml is changed in the backend. When the models.yml is changed, I get an automatic notification that the autoupdate service has to be changed. It is easier for me to auto-generate the files, then to review the changes.

MJJojo97:
Ah ok!

PR #454, opened 2022-04-06
Listen on sig term

### LINKED ISSUE #434: Exit code on normal shutdown
When I run this service in the default docker compose setup provided by the manage tool with `docker-compose up` and I later hit `Ctrl+C` the service (and also all other Golang services) exit with error code `2`. (datastore and auth exit with 137, all other with 0). I would expect 0. What can we do?

ostcar:
@normanjaeckel This fixed #434.

Could you check that this works for the manage-service?

If so, I will create similar PRs for the vote- and icc-service


normanjaeckel:
Looks good. Just merge it and I will see next time ...

normanjaeckel (approved):

PR #460, opened 2022-04-18
Add metrics

ostcar:
This PR adds metrics to the autoupdate-service.

On default, it will print every 15 Minutes a line like this:

```
2022/04/18 08:27:05 Metric: {"connection":{"current":0,"totel":0},"datastore":{"cache_key_len":0,"cache_size":0,"ds_hits":0,"get_calls":0},"runtime":{"goroutines":14}}
```

It starts with the current time, then the prefix "Metric;" and then a json object in one line. The json object will change in the future.

You can configure the metric with the environment variable `METRIC_MINUTES`. Set it to zero to disable metrics.

What do you think. Is 5 minutes a good default? Should the environment variable use MINUTES or should I use SECONDS instead?

Are there values you would like me to add to the metric?

After this is merged and distributed: Please send me the output of some representative instances from time to time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,8 @@ The Service uses the following environment variables:
 * `AUTH_PROTOCOL`: Protocol of the auth servicer. The default is `http`.
 * `OPENSLIDES_DEVELOPMENT`: If set, the service starts, even when secrets (see
   below) are not given. The default is `false`.
+* `METRIC_MINUTES`: Time in minutes how often the metrics are gathered. Zero
+  disables the metrics. The default is `5`.
 
 
 ### Secrets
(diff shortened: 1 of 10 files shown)


peb-adr:
> ```
> 2022/04/18 08:27:05 Metric: {"connection":{"current":0,"totel":0},"datastore":{"cache_key_len":0,"cache_size":0,"ds_hits":0,"get_calls":0},"runtime":{"goroutines":14}}
> ```

typo: 'totel' should be 'total'


> You can configure the metric with the environment variable `METRIC_MINUTES`. Set it to zero to disable metrics.
> 
> What do you think. Is 5 minutes a good default? Should the environment variable use MINUTES or should I use SECONDS instead?

I think personally I would have called it `METRIC_INTERVAL` and have it be set in seconds.


> Are there values you would like me to add to the metric?

I guess time will tell what could be additional interesting values here. On the spot I can't think of more.
Probably when we  start seeing concrete "real-world" problems, we should discuss what values might be of interest to get to the bottom of the problem.

ostcar:
I fixed totel -> total and renamed the environment variable to `METRIC_INTERVAL_SECONDS`. I think it makes it easier when the the name tells you, what kind of value is expected.

PR #474, opened 2022-05-17
Fix corrupt fields in motion

ostcar:
The fireds all_derived_motion_ids and
all_origin_ids can contain ids to motions, that
do not exist. Ignore them


ostcar:
@gsiv This should fix the bug that we talked about. Could you test this the new main of the autoupdate-service with the buggy instance?

PR #488, opened 2022-06-03
Restrict multi at one

ostcar:


ostcar:
@emanuelschuetze this should make the restricter faster. It would be nice, if you could verify this.

But it adds more complexity. So there could be new bugs in it. I think it can be merged. But it would be good, if there is some time before using it in production.

With this PR, some collection can be restricted faster, other can not. It would be good, if we all collections from list views could be fast. Maybe we can talk about it.

ostcar:
I merge this for now so I can continue my work on top of this branch

@emanuelschuetze Maybe don't merge this for a critical instance. It should work. But it is a big change.

PR #502, opened 2022-06-15
Omit empty fields when projecting a poll

ostcar:


internal/projector/slide/poll.go
```diff
@@ -13,8 +13,8 @@ import (
 
 type optionRepr struct {
 	Text            string          `json:"text,omitempty"`
-	ContentObjectID string          `json:"content_object_id"`
-	ContentObject   json.RawMessage `json:"content_object"`
+	ContentObjectID string          `json:"content_object_id,omitempty"`
```
jsangmeister, line 16: `content_object_id` should never be present, even if it is set in the option, since the underlying information is included in `content_object`, so you can remove this line completely (and the code filling it, if any)

ostcar: Unfortunately, this is not possible with the current implementation. The struct is used for decoding the values from the datastore and encoding it for the json output. As far as I know, It is not possible to remove the field only in one of this two cases.

If this is not a problem for the client, I would leave it.  If it is a problem, I would need to solve this by using different structs for decoding and encoding.

jsangmeister: ok, sure. I don't think it is a problem asides from unnecessary data being sent, so we'll leave it as-is.

PR #504, opened 2022-06-16
Try partion updates

ostcar:


ostcar:
This was just a test that is not possible anymore on the current test

PR #518, opened 2022-06-23
New environment variable for ds timeout

ostcar:
This introduces a new environment variable `DATASTORE_TIMEOUT`.

I also changed the environment variable `METRIC_INTERVAL_SECONDS` to `METRIC_INTERVAL`.

Both variables have to flexible unit. Valid values are for example `3s` or `5m`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ The Service uses the following environment variables:
   default is `3s`.
 
 Valid units for duration values are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+One number without a unit is interpreted as seconds. So `3` is the same as `3s`.
 
 ### Secrets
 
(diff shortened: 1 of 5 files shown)


peb-adr:
Ah yes nice. One question though:
Providing a time unit is mandatory now right?
Does the go timespec maybe support falling back to the unit second if only a number was provided?
This would be a bit more lenient/forgiving when you're like "oh I'll just increase the DStimeout real quick" and boom AU is down for a minute because you forgot the unit.

ostcar:
> Ah yes nice. One question though: Providing a time unit is mandatory now right? Does the go timespec maybe support falling back to the unit second if only a number was provided? This would be a bit more lenient/forgiving when you're like "oh I'll just increase the DStimeout real quick" and boom AU is down for a minute because you forgot the unit.

The go function does not support it, but it was easy to implement.

PR #554, opened 2022-10-19
Add API for seach service

ostcar:
@s-l-teichmann This is the required change from the autoupdate-service for the service service.

See the readme how to use it.

Is this ok for you?


s-l-teichmann:
@ostcar 

`curl localhost:9012/internal/autoupdate -d '{"user_id":23,"fqids":["user/1","motion/6"]}'`

Yes. This is the way I would expect it to be. So from my side it seems to be okay.

PR #559, opened 2022-10-22
Log git hash

ostcar:


ostcar:
This does not work in out docker-compose environment

PR #560, opened 2022-10-24
Change motion_comment_section

### LINKED ISSUE #555: Make motion_comment_section read permission implicit for those with write permission
Important for OpenSlides/openslides-client#1709

Autoupdates on the comments in a motion_comment_section should not only be restricted to groups that are contained in the read_group_ids, but also those that are in the write_group_ids

Also take into account the possibility to give special writing (and thereby reading) rights to the motion submitter in accordance with the changes in OpenSlides/openslides-backend#1509 (since there will be at least one new field, that would also have to be tackled):

i.e. expand the motion_comment_section restrictor so that:

- The user has motion.can_see and 
    - has at least one group in common with motion_comment_section/read_group_ids, or
    - has at least one group in common with motion_comment_section/write_group_ids, or
    - submitter_can_write is set to true for this comment section
- Or the user has motion.can_manage.

and change the motion_comment restrictor so that:

- The user has motion.can_see and 
    - has at least one group in common with motion_comment_section/read_group_ids, or
    - has at least one group in common with motion_comment_section/write_group_ids, or
    - submitter_can_write is set to true for the corresponding comment section **AND the user is the submitter of this motion**
- Or the user has motion.can_manage.

ostcar:
Fixes #555

This is a bit different then your description

You wrote for `motion_comment`:
> The user has motion.can_see and ...

I implemented it as
> The user can see the linked motion and ...

This makes a difference in some cases. For example if the user as `motion.can_see` AND he is a submitter of a motion but the motion is in a state, where the submitter can not see it. With your ruleset, he could see the motion_comment but not the motion. With me ruleset, he can not see either.

Is this ok?

### DIFF
--- a/internal/restrict/collection/motion_comment.go
+++ b/internal/restrict/collection/motion_comment.go
@@ -10,7 +10,15 @@ import (
 
 // MotionComment handels restrictions of the collection motion_comment.
 //
-// The user can see a motion comment if the user can see the linked motion and motion comment section.
+// The user can see a motion comment if
+//
+//	The user can see the linked motion and
+//
+//		has at least one group in common with motion_comment_section/read_group_ids, or
+//		has at least one group in common with motion_comment_section/write_group_ids, or
+//		submitter_can_write is set to true for the corresponding comment section AND the user is the submitter of this motion
+//
+//	Or the user has motion.can_manage.
 //
 // Mode A: The user can see the motion comment.
 type MotionComment struct{}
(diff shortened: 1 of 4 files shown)


luisa-beerboom:
> This makes a difference in some cases. For example if the user as `motion.can_see` AND he is a submitter of a motion but the motion is in a state, where the submitter can not see it. With your ruleset, he could see the motion_comment but not the motion. With me ruleset, he can not see either.
> 
> Is this ok?

Yeah, I suppose that's better than what I proposed

luisa-beerboom (approved):
Looks good to me

PR #577, opened 2022-11-13
Add Worker Pool

ostcar:
TODO Add an environment vairable
Write some tests
It does not seem to work. Make sure the pool is used at the right place


ostcar:
I would like to merge this. I see only advantages.

When It is merged, then we can make all other performance tests on top of it.

@rrenkert  Should I just merge it or do you want to test it first?

rrenkert:
Tested this one with about 1000 users and the memory usage is limited in an expected amount.
Please merge :)

rrenkert (approved):

PR #580, opened 2022-11-14
Fix postgres password

ostcar:


peb-adr:
Woah your quick @ostcar 
I wanted to suggest using the other connection string format (key/value) described in
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

I think I heard, it's the preferred format these days. And I think using would have prevented this from happening already.

If your confident, that now all is good, I think you can leave it like this. But maybe the key/value form might be more robust more regards :thinking:

PR #587, opened 2022-11-22
Admins skip the workpool

ostcar:
This skips the workpool for VIPs

In this implementation a VIP is a person that is in a group (any meeting) that has set the attribute `admin_group_for_meeting_id`.

So you have to be a meeting admin for any meeting.

What about organization manager, organization user manager or superadmins?


emanuelschuetze (approved):
Works as expected - with a good performance boost for meeting admins! (e.g. ca. 2s vs >10s waiting time)

PR #590, opened 2022-11-26
restrict use can see performance

ostcar:
Make Restriction faster by precalculate some user ids


ostcar:
@rrenkert ping

rrenkert:
@ostcar Have seen your ping :) I was not able to run performancetests on this since we were busy with the 4.0 release.

ostcar:
Was merged with #605

PR #607, opened 2023-01-14
Add workflow to automatically set project

jsangmeister:


ostcar:
@jsangmeister do you known why this works here?

jsangmeister:
@ostcar interesting, but not really relevant. Runs on other branches than main do not reflect the actual behaviour.

PR #637, opened 2023-03-02
Added a titler for poll_candidate_list objects

### LINKED ISSUE #635: Implement slide content generation for polls with poll_candidate_lists
See OpenSlides/openslides-backend#1608 
There is now a new content_object `poll_candidate_list` for the poll options.

This affects the poll-slide calculation in the autoupdate, which currently returns an error instead of the slide `content` if any of the options contain a poll_candidate-list, since there currently isn't any code for processing an option with a poll_candidate list.

Required behaviour for polls with candidate_list-options:
- The  poll slide data should be generated in exactly the same way as it has been until now, with the exception of the options field.
- The options field is currently being displayed as an array of objects with the form:
```
{
        text:                          string                         // only filled for text options
	content_object_id: string                         // only filled for non-text-options
	content_object:      json.RawMessage   // only filled for non-text-options
	yes:                           string
	no:                            string
	abstain:                    string
	id:                              int
	weight:                     int
}
```

- For text options the fields of the option struct should be filled in the same way as before
- For options with a content_object, as far as I see, there is already code that generates appropriate content for the `ContentObject` field in case the the content_object_id is that of a `user` or a `motion`.  This seems to be done through the `GetTitleInformationFunc` that returns a corresponding `Titler` for current model, which in turn has the `GetTitleInformationFunc` that creates the appropriate object for filling in the `ContentObject` field.
    -> Therefore all that has to be done is register a new kind of titler that handles candidate lists, the rest of the functionality can remain the same.
    -> The object created by this titler should have the following form: 

```
{
    collection: string = `poll_candidate_list`
    content_object_id: string //The fqid of the poll_candidate_list
}
```

This could look like this:

```
POLL CANDIDATE LIST:
{
    id: 1
    poll_candidate_ids: [1, 2]
    ...
}
```

would become:
```
{
    collection: `poll_candidate_list`
    content_object_id: `poll_candidate_list/1`
}
```

This will probably be enough since the finished slide isn't supposed to display the entries of the candidate list, just the poll results

luisa-beerboom:
Closes #635


ostcar (approved):
I have not tested this, but the Code looks good. 

Please keep in mind that the way how to access the database in the projector package is a bit strange. In other parts of the service it is handled differently.

ostcar:
There is one formatting error. But the github action does not show it :(

Could you run `go fmt ./...` in the branch? This should fix it.

luisa-beerboom:
> There is one formatting error. But the github action does not show it :(
> 
> Could you run `go fmt ./...` in the branch? This should fix it.

@ostcar  I would, but I currently can't directly run any go commands in my setup, probably because of an old golang version. If I don't get this fixed by Friday evening, could you do the formatting?

PR #659, opened 2023-04-06
RestrictFQIDs add field filter param

bastianjoel:
This is needed for the search service.


internal/autoupdate/autoupdate.go
```diff
@@ -263,7 +263,7 @@ func (a *Autoupdate) HistoryInformation(ctx context.Context, uid int, fqid strin
 // list of fqids.
 //
 // The return format is a map from fqid to an object as map from field to value.
-func (a *Autoupdate) RestrictFQIDs(ctx context.Context, userID int, fqids []string) (map[string]map[string][]byte, error) {
+func (a *Autoupdate) RestrictFQIDs(ctx context.Context, userID int, fqids []string, requestedFields map[string]map[string]struct{}) (map[string]map[string][]byte, error) {
```
ostcar, line 266: I don't understand the function. There is the attribute `fqids` and there is the attribute `requestedFields`. Are they distinct? Would it be possible to remove the attribute `fqids` and just use requtedFields? Or maybe another attribute like a list of `fqfields`?

If not, it is ok for me. But in this case, please document the meaning of the attributes in the docstring

internal/autoupdate/autoupdate.go
```diff
@@ -282,8 +282,13 @@ func (a *Autoupdate) RestrictFQIDs(ctx context.Context, userID int, fqids []stri
...
 		for _, field := range fields {
-			key := dskey.Key{Collection: collection, ID: id, Field: field}
-			keys = append(keys, key)
+			if _, ok := requestedFields[collection]; !ok {
+				key := dskey.Key{Collection: collection, ID: id, Field: field}
+				keys = append(keys, key)
+			} else if _, ok := requestedFields[collection][field]; ok {
+				key := dskey.Key{Collection: collection, ID: id, Field: field}
+				keys = append(keys, key)
+			}
```
ostcar, line 291: This looks strange. There are two conditions with the same two lines of code in it. It says: "if the collection was not requested, return it, if the field of the collection was requested, return it". Is this really what you want? If so, please document it.

The code could be clearer like this:

```go
_, hasCollection := requestedFields[collection]
hasCollectionField := false
if hasCollection {
  _, hasCollectionField = requestedFields[collection][field]
}

if !hasCollection || hasCollectionField {
  key := dskey.Key{Collection: collection, ID: id, Field: field}
  keys = append(keys, key)
}
```

But if possible, I would recommend not to use this code, but to make the function signature easier. For example only request a list of fqFields.

internal/autoupdate/autoupdate_test.go
```diff
@@ -122,10 +122,11 @@ func TestRestrictFQIDs(t *testing.T) {
 		user/1:
 			username: superadmin
 			first_name: kevin
+			last_name: foo
 	`))
 	s, _, _ := autoupdate.New(environment.ForTests{}, ds, RestrictAllowed)
 
-	got, err := s.RestrictFQIDs(ctx, 1, []string{"user/1"})
+	got, err := s.RestrictFQIDs(ctx, 1, []string{"user/1"}, map[string]map[string]struct{}{"user": {"id": {}, "username": {}, "first_name": {}}})
```
ostcar, line 129: It seems, that your current implementation has more then one case. You should test the other cases. For example:

```go
s.RestrictFQIDs(ctx, 1, []string{"user/1"}, nil) // Returns all fields of user
s.RestrictFQIDs(ctx, 1, []string{"user/1"}, map[string]map[string]struct{}{"group": nil}) // Returns all fields of user
s.RestrictFQIDs(ctx, 1, []string{"user/1"}, map[string]map[string]struct{}{"user": nil }) // Returns no data
s.RestrictFQIDs(ctx, 1, []string{"user/1"}, map[string]map[string]struct{}{"user": {"username": {}}}) // Returns only user/1/username
```

internal/http/http.go
```diff
@@ -254,8 +254,9 @@ func HandleRestrictFQIDs(mux *http.ServeMux, service restrictFQIDser) {
 		prefixInternal+"/restrict_fqids",
 		func(w http.ResponseWriter, r *http.Request) {
 			var requestBody struct {
-				UserID int      `json:"user_id"`
-				FQIDs  []string `json:"fqids"`
+				UserID int                 `json:"user_id"`
+				FQIDs  []string            `json:"fqids"`
+				Fields map[string][]string `json:"fields"`
```
ostcar, line 259: Here it is the same as above. I don't understand the requirement from the search service. But it seems strange, that it requests a list of fqids and a list of fields. Does the search service know witch fields it wants? if so, it could just request them.

internal/http/http.go
```diff
@@ -267,7 +268,15 @@ func HandleRestrictFQIDs(mux *http.ServeMux, service restrictFQIDser) {
 				return
 			}
 
-			restricted, err := service.RestrictFQIDs(r.Context(), requestBody.UserID, requestBody.FQIDs)
+			requestedFieldsMap := make(map[string]map[string]struct{}, len(requestBody.Fields))
```
ostcar, line 271: If you want to use a set, you could use this generic set package, I wrote for the autoupdate service: https://github.com/OpenSlides/openslides-autoupdate-service/blob/main/pkg/set/set.go

```go
requestedFieldsMap := make(map[string]set.Set[string], len(requestBody.Fields)) 
```

(I have not tested this)

ostcar (commented):
Some remarks.

Since all code changes are only on function that are only used by the search service, my remarks are not important. It would be nice, if you could document what the functions does in the doc strings.

If you want to merge this to keep going with the search service, this is ok for me

internal/autoupdate/autoupdate.go
bastianjoel: `requestedFields` is a map of collection names containing fields. E.g.: `{ user: [id, firstname] }`.

ostcar (approved):
Have not tested this. But looks good

PR #670, opened 2023-04-23
Set keep alive header

ostcar:


ostcar:
Could be a problem. See  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection

ostcar:
@emanuelschuetze I don't think this will help. So I close it for now

PR #671, opened 2023-04-24
Connection count

ostcar:


ostcar:
@peb-adr This is a new feature to count the current connections. @emanuelschuetze requested this.

to fetch the data, you can call ./openslides-performance request /system/autoupdate/connection_count

Maybe the client can integrate it somewhere. Or you can integrate it in graphana?

ostcar:
The values should be logged with the metrics.

ostcar:
As discussed:

Each autoupdate-service has its own counter in memory.
When it starts, it generates its own random value (maybe uuid).
When a request is open or closed, it saves its own counter with the current timestamp and its uuid to redis.
To get the current value, all counters in the redis are comulated, but only for values, where the timestamp is not to old.
When there are no new connections, each services updates its counter in redis by updating the timestamp.

ostcar:
Closed in favor of https://github.com/OpenSlides/openslides-autoupdate-service/pull/692

PR #674, opened 2023-04-29
Lead motion

ostcar:


ostcar:
@luisa-beerboom and @jsangmeister this checks for the permission of lead motions.

It was harder then I thought. The problem was, that I wanted this code also to work, if there is a circle in the lead motions. (motion1 -> motion2 -> motion3 -> motion1 ...). I know, that the backend will probably prevent this. But my first implementation went into an endless loop. I think the code should not do this, even if the database is corrupted.

rrenkert (approved):

PR #678, opened 2023-05-08
Changed motion slide to show motion text in amendments regardless of …

luisa-beerboom:
…settings

Should fix the remark in [this client change request](https://github.com/OpenSlides/openslides-client/pull/2258#pullrequestreview-1414677992) on the PR for issue Openslides/openslides-client#2218

### DIFF
--- a/internal/projector/slide/motion.go
+++ b/internal/projector/slide/motion.go
@@ -169,10 +169,9 @@ func Motion(store *projector.SlideStore) {
 			"recommendation_id",
 			"recommendation_extension",
 			"recommendation_extension_reference_ids",
+			"text",
 		}
-		if meeting.MotionsEnableTextOnProjector {
-			fetchFields = append(fetchFields, "text")
-		}
+
 		if meeting.MotionsEnableReasonOnProjector {
 			fetchFields = append(fetchFields, "reason")
 		}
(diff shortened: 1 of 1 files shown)


ostcar (approved):

ostcar:
I have not test this. The code looks good. It should do, what the description does.

Please keep in mind, that this is a permission thing. This this PR, each user with the permission to see the projector gets the motion text from all motions that are currently on the projector OR where on the projector in the past.

The setting [meeting/motions_enable_text_on_projector](https://github.com/OpenSlides/openslides-backend/blob/main/global/meta/models.yml#L1069) will be only client-side. The server will return the motion text regardless of this setting.

luisa-beerboom:
@ostcar The point of this PR was to
1. Always show the text if the motion is an amendment (i.e. there is a `lead_motion`), but 
2. Continue using the [meeting/motions_enable_text_on_projector](https://github.com/OpenSlides/openslides-backend/blob/main/global/meta/models.yml#L1069) setting to decide whether to show the text for normal motions

As far as I see it, that doesn't strictly make the setting "only client-side", but just lessens its application area in a way that honestly makes sense IMO (since one can also already see the diff view for the diff amendments, one should probably be able to see the text for the text amendments).

Your comment that this PR will make the setting only client-side confuses me, especially since I am not yet that experienced with go. Am I somehow not doing what I set out to do? If I am not, please make a change request.

ostcar:
> Your comment that this PR will make the setting only client-side confuses me, especially since I am not yet that experienced with go. Am I somehow not doing what I set out to do? If I am not, please make a change request.

Your are right and I am sorry. I did check your code again and you are 100% correct.

ostcar (approved):

luisa-beerboom:
Okay, I misunderstood the assignment, it seems. Closing this now

PR #683, opened 2023-05-10
Remove content from non current projections

### LINKED ISSUE #679: Reduce autoupdates for projections
Currently, all users which have the permission `projector.can_see` get autoupdates for all projection in the preview or history of all projectors. This is unnecessary or even unwanted data being sent to all such users, since e.g. all motions which have been projected at one point in time will be sent to all users if the motion text changes.

Proposed changes:
- projections in the preview or history of a projector (so, without `current_projector_id` set) should only be sent to users with `projector.can_manage`
- the `content` field of such projections can be reduced to only the necessary data, since the client does not display the full projection in the preview/history, but only the title. @bastianjoel @luisa-beerboom maybe you can give more input here which fields are needed for preview/history, if necessary on a per-collection basis.

ostcar:
Fixes #679

### DIFF
--- a/pkg/datastore/datastore.go
+++ b/pkg/datastore/datastore.go
@@ -62,6 +62,7 @@ type Datastore struct {
 	changeListeners  []func(map[dskey.Key][]byte) error
 	calculatedFields map[string]func(ctx context.Context, key dskey.Key, changed map[dskey.Key][]byte) ([]byte, error)
 	calculatedKeys   map[dskey.Key]string
+	calculatedKeysMu sync.RWMutex
 
 	history HistoryInformationer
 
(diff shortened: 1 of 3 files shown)


pkg/datastore/datastore.go
bastianjoel, line 1: Was this supposed to be added to this PR? I don't exactly see how this is related to the original issue.

bastianjoel (approved):

pkg/datastore/datastore.go
ostcar: Yes and No. There was a race condition. The test did not catch it before. But now, the function to generate projection/content is much faster (when `current_projection_id` is `nil`) so the race condition did happen.

So the bug was in the code before and could be fixed in a separate PR, but it is only relevant to fix it for this PR

PR #690, opened 2023-05-24
Fix bug in projector

ostcar:
The projector did try to fetch its own value from the cache. This did not work, when the cache was cleared (every 24 minutes).

### DIFF
--- a/environment.md
+++ b/environment.md
@@ -27,6 +27,7 @@ The Service uses the following environment variables:
 * `AUTH_PORT`: Port of the auth service. The default is `9004`.
 * `AUTH_FAKE`: Use user id 1 for every request. Ignores all other auth environment variables. The default is `false`.
 * `CONCURENT_WORKER`: Amount of clients that calculate there values at the same time. Default to GOMAXPROCS. The default is `0`.
+* `CACHE_RESET`: Time to reset the cache. The default is `24h`.
 * `METRIC_INTERVAL`: Time in how often the metrics are gathered. Zero disables the metrics. The default is `5m`.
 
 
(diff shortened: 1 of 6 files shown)


ostcar:
Currently, there is no environment variable to clear the cache.

I don't think, that we should add it just to test this PR. But if you need it, I could add it.

Actually, I would like to remove the cache clear feature at all. It is hacky and could have other problems like this one.  The only downside would be, that the autoupdate service could use more memory when it is running for a long time. But I think it would be clearer and more effective, to restart it from time to time.

@peb-adr do you think I could remove the feature?

ostcar:
@peb-adr I added the environment variable `CACHE_RESET`. You can test this by setting a value like `5s`

After we merge this: Please test this with a very high value like `720h` So we can test, if the reset feature could be removed.

peb-adr:
> Actually, I would like to remove the cache clear feature at all. It is hacky and could have other problems like this one. The only downside would be, that the autoupdate service could use more memory when it is running for a long time. But I think it would be clearer and more effective, to restart it from time to time.
> 
> @peb-adr do you think I could remove the feature?

Hmmm. To me it doesn't sound desirable to remove it. 
I couldn't say how much of an instances data is in cache after 24h (obviously this also depends on the users activity). But for our big instances i would estimate it can be <5%.
So I would be afraid of how much it may grow when all data makes it into cache.

Wouldn't it be doable to have data in cache be cleared on a per-key basis after a set time?
I don't know if that would be less hacky though ...
Regardless: I am not a fan of having to restart the AU regularly!

However I think this discussion should go into a separate issue.


The concrete problem addressed in this PR is solved. I just recreated it on `main` with reduced cacheResetTime.
-> logs when trying to see an new projection after cacheResetTime:
```
Error calculating key projection/6/content: getting old value: getOrSet`: fetching missing keys: waiting for fetch missing: context deadline exceeded
```
With this PR and `CACHE_RESET=2m` the faulty behavior for projections doesn't occur.

PR #692, opened 2023-06-03
Connection count v2

ostcar:


ostcar:
TODO: Add an Environment Variable to disable connection count

ostcar:
@m-schieder I added the environment variable to disable the vote count. See  https://github.com/OpenSlides/openslides-autoupdate-service/pull/692/files#diff-1784e41ee1e7db27f3848b8870719437db12f9c8dcc552d35ed9e32262f8e367R33

I also tested it with

```
echo "request: GET https://localhost:8000/system/autoupdate?single=1 []" | ./openslides-performance browser replay -n 10000 --close
```

I takes around 2,4 Seconds to open the 10.000 empty requests to the autoupdate service. This does not change if the metric is enabled or disabled.

ostcar:
Could we merge it first? With the new environment variable, it can be disabled anyway

internal/http/connection_count.go
```diff
@@ -0,0 +1,189 @@
...
+func (c *connectionCount) Metric(con metric.Container) {
+	ctx := context.Background()
+
+	data, err := c.metric.Get(ctx)
+	if err != nil {
+		oserror.Handle(fmt.Errorf("fetch connection count metric from redis: %w", err))
+		return
+	}
+
+	localCurrenedUsers := 0
```
bastianjoel, line 117: Variable name spelling seems not correct.

pkg/redis/metric.go
```diff
@@ -0,0 +1,134 @@
...
+func buildInstanceID(now func() time.Time) string {
+	t := now().UTC().Format("2006-01-02T15:04:05")
+
+	const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	const length = 8
+
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[rand.Intn(len(charset))]
+	}
```
bastianjoel, line 59: Maybe use https://pkg.go.dev/crypto/rand#Read

bastianjoel (approved):
Except for the mentioned performance concerns this looks good to me.

pkg/redis/metric.go
ostcar: The main difference between `math/rand` and `crypto/rand` is, that the one is pseudo random and the other uses /dev/rand.

This random string is only necessary, so different instances get different names. Why do you think they should be crypto random?

m-schieder:
> Could we merge it first? With the new environment variable, it can be disabled anyway

I don't like dead source code that much, but if the feature is turned off by default so the self-hosters and we don't have to set an extra environment variable, that's fine with me.

pkg/redis/metric.go
bastianjoel: `rand.Read` just feels more convenient to use as you don't need to define the alphabet before. I am fine with either.

ostcar:
Do not save the values on each request but at a interval

pkg/redis/metric.go
ostcar: rand.Read returns raw bytes (beween 0 and 255). So you still have to define what characters you want to use

ostcar:
> Do not save the values on each request but at a interval

Done. Can we merge it now?

bastianjoel:
Much better now
```
File: openslides-autoupdate-service
Type: cpu
Time: Jul 3, 2023 at 10:04am (CEST)
Duration: 15.01s, Total samples = 550ms ( 3.67%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 360ms, 65.45% of 550ms total
Showing top 10 nodes out of 192
      flat  flat%   sum%        cum   cum%
     130ms 23.64% 23.64%      130ms 23.64%  runtime/internal/syscall.Syscall6
      90ms 16.36% 40.00%       90ms 16.36%  runtime.futex
      40ms  7.27% 47.27%       40ms  7.27%  runtime.nanotime
      30ms  5.45% 52.73%       30ms  5.45%  runtime.osyield
      20ms  3.64% 56.36%       20ms  3.64%  time.Now
      10ms  1.82% 58.18%       10ms  1.82%  crypto/hmac.New
      10ms  1.82% 60.00%       50ms  9.09%  encoding/json.(*decodeState).object
      10ms  1.82% 61.82%       10ms  1.82%  encoding/json.stateBeginValue
      10ms  1.82% 63.64%       90ms 16.36%  github.com/OpenSlides/openslides-autoupdate-service/internal/autoupdate.(*connection).Next.func1
      10ms  1.82% 65.45%       30ms  5.45%  internal/poll.accept
```

bastianjoel (approved):

m-schieder (approved):

PR #705, opened 2023-06-22
Implement new WiFi access data slide

### LINKED ISSUE #704: New meeting slide type for WiFi access data
It only needs to deliver the given meeting's WiFi settings (`users_pdf_wlan_ssid`, `users_pdf_wlan_password`, `users_pdf_wlan_encryption`).

luisa-beerboom:
Closes #704 

@rrenkert This is necessary for the upcoming PR for OpenSlides/openslides-client#2122

### DIFF
--- a/internal/projector/slide/meeting.go
+++ b/internal/projector/slide/meeting.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
+	"strings"
 
+	"github.com/OpenSlides/openslides-autoupdate-service/internal/projector"
 	"github.com/OpenSlides/openslides-autoupdate-service/internal/projector/datastore"
 )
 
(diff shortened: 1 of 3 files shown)


internal/projector/slide/meeting.go
```diff
@@ -23,6 +26,9 @@ type dbMeeting struct {
 	ListOfSpeakersAmountNextOnProjector       int    `json:"list_of_speakers_amount_next_on_projector"`
 	ListOfSpeakersAmountLastOnProjector       int    `json:"list_of_speakers_amount_last_on_projector"`
 	ListOfSpeakersShowAmountOfSpeakersOnSlide bool   `json:"list_of_speakers_show_amount_of_speakers_on_slide"`
+	UsersPdfWlanSsid                          string `json:"users_pdf_wlan_ssid"`
```
ostcar, line 29: You can keep it that way. But there is a convention in Go to write all letters of an acronym in upper letters: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

So you could rename this to `UsersPDFWLANSSID` But I see, that this looks strange in this example :)

internal/projector/slide/meeting.go
```diff
@@ -50,3 +56,35 @@ func getMeeting(ctx context.Context, fetch *datastore.Fetcher, meetingID int, fe
 	}
 	return meeting, nil
 }
+
+// WiFiAccessData renders the wifi_access_data slide.
+func WiFiAccessData(store *projector.SlideStore) {
+	store.RegisterSliderFunc("wifi_access_data", func(ctx context.Context, fetch *datastore.Fetcher, p7on *projector.Projection) (encoded []byte, err error) {
+		meetingID, err := strconv.Atoi(strings.Split(p7on.ContentObjectID, "/")[1])
+		if err != nil {
+			return nil, fmt.Errorf("error in Atoi with ContentObjectID: %w", err)
```
ostcar, line 65: The error messages will be joint. So do not start any error message with `error`. Or the full error message will look like: `Error: error in function1: error in function2: error in function3`. Instead just write, what you tried to do. For example: `convert string to int: %w`. So the full error message will look like: `Error: function1: function2: function3`

ostcar (approved):
Two proposals. If you like, you can merge it as it is.

internal/projector/slide/meeting.go
luisa-beerboom: I'll do it anyway

PR #723, opened 2023-07-12
Histery package

ostcar:
This removes the history functionally from the datastore package and puts it into its own package.

This is a required step for the restrictor-refactor-project.


ostcar:
* [ ] test it

PR #727, opened 2023-07-19
Feature/vote decrypt

ostcar:


ostcar:
We have a new concept. This PR is not needed anymore. Where will be a new PR

PR #733, opened 2023-07-26
Pre restrict

ostcar:


internal/restrict2/restrict_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/OpenSlides/openslides-autoupdate-service/pkg/datastore/dsmock"
+)
+
+func TestRestritGet(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	flow := dsmock.NewFlow(
+		dsmock.YAMLData(`---
+		user/1/group_$_ids: ["5"]
```
r-peschke, line 19: Why are you still using template fields?

ostcar: I created this PR before I found the performance gains by using another key-type. At this point, we though, that we had to do last weeks congress with the template fields. Emanuel ask be to rewrite the restricter on the old main branch.

After I changed the key-type, we merged the rtf-branch.

Now I have to redo all the work here on the current main. I probably will create a new branch for that :)

PR #761, opened 2023-08-27
Connection count for admin

### LINKED ISSUE #749: Make connection_count route also work for meeting admins
Meeting admins should be allowed to call the connection_count route.
The return value for them should only include the users in the meetings where they are admin

ostcar:
Fixes #749

This should be solve it.

With this implementation, a meeting admin is not allowed to see the connections from anonymous users. This was easier to test. We can change this in the future, if needed.


luisa-beerboom (approved):
Can't really comment on the code itself, but I tried it out with a modified client and it works wonderfully as far as I can see

luisa-beerboom:
If this is done, please merge

PR #770, opened 2023-09-11
Move profiler to internal route

### LINKED ISSUE #765: Change profile routes to be internal
We want to gather some profile data. But the fact it is atm publicly served when enabled is concerning.

Please change the routes to be internal such that it will not get routed by the proxy.
We will then get the profiles by `curl`ing them from inside the stack or something similar.

bastianjoel:
resolves #765


ostcar:
I would propose, that we use the go default routes for profiling. I only used the special routes, so they could be accessed from extern. But since we don't want that, the normal routes can be used

ostcar:
It would be nice, if you wait for my approval in the services, that I maintain.

I will add the profiling routes to the vote and icc service and change them in the autoupdate service as well

PR #776, opened 2023-09-20
Reset cache on projector reset

### LINKED ISSUE #775: Projector content not updated after cache reset
After a cache reset the projector content of e.g. countdown do not get updated anymore.

bastianjoel:
resolves #775

### DIFF
--- a/internal/projector/projector.go
+++ b/internal/projector/projector.go
@@ -54,6 +54,7 @@ func (p *Projector) Reset() {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	p.cache = make(map[dskey.Key][]byte)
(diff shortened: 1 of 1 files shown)


ostcar (changes requested):
I think, you have to reset the cache AND the hotkeys.

It should not be a problem, if there are old Data in the hotkeys. When the cache is empty, then the hotkeys will be regenerated as soon, as the projection is calculated for the next time. But if the projection is never called after the reset, then the hotkeys will never be cleared. So I think it is better to clear them in the reset call.

ostcar (approved):
I have not tested it.

m-schieder (approved):

PR #781, opened 2023-09-29
Add restrictor for import_preview

ostcar:
@r-peschke can you tell me, who can see the new collection `import_preview`?

@ostcar It is backend internal. The client receives the preview data with the response of the type specific json_upload-action. In the payload to import the data the client uses the `id`  of the `import_preview` instance.

IMO you should restrict the whole collection always.

### DIFF
--- a/environment.md
+++ b/environment.md
@@ -33,4 +33,3 @@ The Service uses the following environment variables:
 * `METRIC_INTERVAL`: Time in how often the metrics are gathered. Zero disables the metrics. The default is `5m`.
 * `METRIC_SAVE_INTERVAL`: Interval, how often the metric should be saved to redis. Redis will ignore entries, that are twice at old then the save interval. The default is `5m`.
 * `DISABLE_CONNECTION_COUNT`: Do not count connections. The default is `false`.
-* `ENABLE_PROFILE_ROUTES`: Activate development routes for profiling. The default is `false`.
--- a/internal/restrict/collection/collection.go
+++ b/internal/restrict/collection/collection.go
@@ -170,6 +170,7 @@ var collectionMap = map[string]Restricter{
 	ChatMessage{}.Name():                ChatMessage{},
 	Committee{}.Name():                  Committee{},
 	Group{}.Name():                      Group{},
+	ImportPreview{}.Name():              ImportPreview{},
 	Mediafile{}.Name():                  Mediafile{},
 	Meeting{}.Name():                    Meeting{},
 	MeetingUser{}.Name():                MeetingUser{},
--- a/internal/restrict/collection/import_preview.go
+++ b/internal/restrict/collection/import_preview.go
@@ -2,16 +2,13 @@ package collection
 
 import (
 	"context"
-	"errors"
 
 	"github.com/OpenSlides/openslides-autoupdate-service/pkg/datastore/dsfetch"
 )
 
 // ImportPreview handels restrictions of the collection import_preview.
 //
-// The user can see a import_preview, if TODO
-//
-// Mode A: The user can see the import_preview.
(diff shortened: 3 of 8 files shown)


jsangmeister:
`import_preview/name` contains the import type. The user needs the respective manage permissions to execute the import, so I think he should also need them to see the preview. The list of current and planned permissions would therefore be:
- `topic`: `agenda_item.can_manage`
- `account`: OML `can_manage_users`
- `participant`: `user.can_manage`
- `committee`: OML `can_manage_organization`
- `motion`: `motion.can_manage`

For the meeting permissions, you will need the meeting id. You could either extract it from the import data, which would require the AU service to need knowledge about the structure of this data (not a fan), or the backend needs to save it in an additional field `meeting_id`, which would be empty for the organization imports. Also, we should add  an enum restricting the values of `import_preview/name` (also, `type` might be more appropriate than `name` then).

@r-peschke @ostcar what do you think?

ostcar:
Who can start the import? If only orga-managers (of some level) can do it, I could use the orga permission.

If a meeting admin can start the import, then I would need the meeting_id (please in a separate field) and only let the meeting admin see the data.

You could also add the user_id of the user, that started the import. Then it would be possible, that only this user (and the superadmin) can see the collection

r-peschke:
The client don't need to see the import_preview-collection!
The preview data is passed in the request response to the client.
The collection is, like the action_worker-collection, only used for internal use of the backend (transfer data from json_upload-action to be used in import-action.

> `import_preview/name` contains the import type. The user needs the respective manage permissions to execute the import, so I think he should also need them to see the preview. The list of current and planned permissions would therefore be:
> 
>     * `topic`: `agenda_item.can_manage`
> 
>     * `account`: OML `can_manage_users`
> 
>     * `participant`: `user.can_manage`
> 
>     * `committee`: OML `can_manage_organization`
> 
>     * `motion`: `motion.can_manage`
> 
> 
The AU don't need this collection at all!
> For the meeting permissions, you will need the meeting id. You could either extract it from the import data, which would require the AU service to need knowledge about the structure of this data (not a fan), or the backend needs to save it in an additional field `meeting_id`, which would be empty for the organization imports. Also, we should add an enum restricting the values of `import_preview/name` (also, `type` might be more appropriate than `name` then).
> 
> @r-peschke @ostcar what do you think?

see above.

jsangmeister:
Right, my bad! I still think the name/type should be constrained by an enum to ensure valid types, but I'll open a backend issue for that.

ostcar:
The autoupdate-service can not ignore a collection from the model.yml. The models.yml does not support such a ignore feature.

In my opinion, we have the models.yml to define the data needed by the backend, autoupdate and client. If the `import_preview` is only used as same sort of cache for the backend, then the question is, if the models.yml is the right place.

The action_worker-collection is not a similar case, since the client gets the backend responce via the autoupdate-service.

I added a restriction, so nobody (except the superadmin) can see the collection. This works.

ostcar:
> > I added a restriction, so nobody (except the superadmin) can see the collection. This works.
> 
> Ok, but at the moment nobody needs this permission, even the superadmin.

So why do you send changes to the collection on the message bus?

r-peschke:
> > > I added a restriction, so nobody (except the superadmin) can see the collection. This works.
> > 
> > 
> > Ok, but at the moment nobody needs this permission, even the superadmin.
> 
> So why do you send changes to the collection on the message bus?

Because the datastore base send it for all changes. Why do we should make an exception? At the moment there is nobody with a subscription.

jsangmeister:
Additionally, the datastore has no knowledge of the different collections, so we would have to build a way to differentiate between `import_preview` and other collections.

PR #814, opened 2024-01-12
LoS extension projector

bastianjoel:


ostcar (approved):
Looks good.

Could you also remove the user slide? Or do you want to remove it in a separate PR?

bastianjoel:
I would prefer doing this in another PR.

PR #817, opened 2024-01-18
Add structure level slides

bastianjoel:


ostcar (approved):
I don't like the code, but this is not your fold. Its this ugly API 😁

Issue #816, opened 2024-01-18
Add new motion restriction modes


jsangmeister:
see https://github.com/OpenSlides/openslides-backend/issues/2119:

- [x] The new motion restriction mode B should only be visible to users with `motion.can_manage`

see https://github.com/OpenSlides/openslides-backend/issues/2120:

- [ ] TODO

jsangmeister:
For https://github.com/OpenSlides/openslides-backend/issues/2120, we still need to decide on how to exactly model it. A possible way could be to implement this in the restricter the following way: To determine if the user can see the field `motion/recommendation_id`, the restricter follow the relation and checks the field `motion_state/internal_recommendation`. If that is true, the user needs `motion.can_manage` to see the recommendation. Otherwise, restriction mode C is used. @ostcar will evaluate how feasible this solution is or if there is a better way to model it.

PR #819, opened 2024-01-19
Add new motion fields

jsangmeister:
part of https://github.com/OpenSlides/openslides-autoupdate-service/issues/816


internal/restrict/collection/motion.go
```diff
@@ -115,6 +119,20 @@ func (m Motion) see(ctx context.Context, ds *dsfetch.Fetch, motionIDs ...int) ([
 	})
 }
 
+func (m Motion) canManage(ctx context.Context, ds *dsfetch.Fetch, motionIDs ...int) ([]int, error) {
```
ostcar, line 122: Could you rename this to `modeB`? Maybe my naming of this functions is not the best. But I do it for all mods for all collections the same. So it would like to have a uniform system.

internal/restrict/collection/motion.go
```diff
@@ -115,6 +119,20 @@ func (m Motion) see(ctx context.Context, ds *dsfetch.Fetch, motionIDs ...int) ([
...
+	return eachMeeting(ctx, ds, m, motionIDs, func(meetingID int, ids []int) ([]int, error) {
```
ostcar, line 123: I have a helper function, that checks, if the user has a permission. Please use this: https://github.com/OpenSlides/openslides-autoupdate-service/blob/a03f851e222cd63112cf6ffe45e86e277b021741/internal/restrict/collection/collection.go#L267-L279

ostcar (changes requested):

ostcar:
I now, this is only a draft. Please add tests for the new restrictor, when it is ready to merge:

https://github.com/OpenSlides/openslides-autoupdate-service/blob/main/internal/restrict/collection/motion_test.go

internal/restrict/collection/motion.go
jsangmeister: Sure

internal/restrict/collection/motion.go
jsangmeister: thanks for the hint!

jsangmeister:
I added some rudimentary tests which I think suffice for this simple feature. If you want more or other tests, please specifiy what should be tested, as I do not have enought insight in the test setup.

ostcar:
Looks perfect. For this simple case, the test does not seem necessary. But it helps me when refactoring the restricted.

PR #824, opened 2024-01-26
Add identical motions

jsangmeister:
updates files according to https://github.com/OpenSlides/openslides-backend/issues/2121


jsangmeister:
I don't think we should merge AU PRs when the related backend PR is still open. It leads to false positive model changes like this: https://github.com/OpenSlides/openslides-autoupdate-service/pull/827

ostcar:
How is this possible? The auto generated PR is only created, when the backend PR is merged into main

jsangmeister:
Yes, but this was not an auto-generated PR, as the title and the author would also suggest ;) I created this PR to have a working dev setup available in conjunction with the backend PR to be able to implement the client side.

PR #833, opened 2024-02-03
Long polling fallback

ostcar:


bastianjoel (changes requested):
I cannot properly work with the multipart response without the boundary set in the `Content-type` header. 
It would also be easier (and probably more safe) to handle the data if named form fields were used for the response.

bastianjoel:
Something like this would work: https://github.com/bastianjoel/openslides-autoupdate-service/commit/ab566d964d0985828432a324e9293c98a6f47fce

ostcar:
@bastianjoel the easiest way to solve a `go.sum` conflict is by deleting the file and calling `go mod tidy` afterwards.

bastianjoel:
The `Content-Type` header is still set as `multipart` if an error occurs. This still needs to be changed.

bastianjoel (approved):

PR #842, opened 2024-02-08
Update meta

peb-adr:
paths changed for meta submodule.
Reflect these changes in order to be able to merge `main -> feature/los-extension`

Also called `go generate ./...` which yielded no changes. This commit is therefore safe IMO.


peb-adr:
might be related to  #841

jsangmeister:
> might be related to #841

It is the same, except that I also adjusted the docs ;) I'm therefore closing this as it is superseded by #841

PR #843, opened 2024-02-09
Unify name of the openslides-meta submodule

peb-adr:
... with respect to all other existent submodules referenced from the main repo


jsangmeister (approved):
Do we have to run any command locally after this is merged so that git picks up on the change? I presume that this does not happen automatically...

peb-adr:
> Do we have to run any command locally after this is merged so that git picks up on the change? I presume that this does not happen automatically...

Oh, good remark. Yes a `git submodule update --init` will be required to instruct git to reread `.gitmodules` and register the submodule on the (already checked out) path.

PR #844, opened 2024-02-12
Enhance motion fields

jsangmeister:
needed for https://github.com/OpenSlides/openslides-client/pull/3164


jsangmeister:
Interestingly, the new feature works even without these changes. Does the autoupdate service simply pipe through unkown collections?

bastianjoel:
So a motion with state restriction `is_submitter` should be visible to a user if the user is an editor? Is this correct? @emanuelschuetze

jsangmeister:
> So a motion with state restriction `is_submitter` should be visible to a user if the user is an editor? Is this correct? @emanuelschuetze

talked with @emanuelschuetze: No, there should be no changes to the visibility of a motion for editors/working group speakers

internal/restrict/collection/motion_editor.go
```diff
@@ -0,0 +1,46 @@
...
+
+// MotionEditor handels restrictions of the collection motion_editor.
+//
+// The user can see a motion_editor level if he has `motion.can_see`
+//
+// Mode A: The user can see the motion editor.
+type MotionEditor struct{}
+
+// Name returns the collection name.
+func (e MotionEditor) Name() string {
```
ostcar, line 19: Could you change the `e` parameter to `m`. I always use the first letter of the collection

internal/restrict/collection/motion_editor.go
```diff
@@ -0,0 +1,46 @@
...
+
+	return meetingID, true, nil
+}
+
+// Modes returns the restrictions modes for the meeting collection.
+func (e MotionEditor) Modes(mode string) FieldRestricter {
+	switch mode {
+	case "A":
+		return e.see
+	case "B":
```
ostcar, line 38: There is no collection mode `B`. Please remove this case

internal/restrict/collection/motion_working_group_speaker.go
```diff
@@ -0,0 +1,46 @@
...
+
+	return meetingID, true, nil
+}
+
+// Modes returns the restrictions modes for the meeting collection.
+func (m MotionWorkingGroupSpeaker) Modes(mode string) FieldRestricter {
+	switch mode {
+	case "A":
+		return m.see
+	case "B":
```
ostcar, line 38: There is no collection mode `B`. Please remove this case

ostcar (changes requested):
This should probably be a draft until the changes to meta was merged into its main branch.

If there is a PR (not a draft PR) that looks good, then I think, that I should merge it.

ostcar (approved):

emanuelschuetze (dismissed):
I cannot test this PR (together with client and backend). There is an error after trying to login with a normal user (superadmin works)!

```
autoupdate-1        | 2024/02/15 06:52:33 http: panic serving 172.21.0.14:36924: runtime error: invalid memory address or nil pointer dereference
```

ostcar:
> I cannot test this PR (together with client and backend). There is an error after trying to login with a normal user (superadmin works)!
> 
> ```
> autoupdate-1        | 2024/02/15 06:52:33 http: panic serving 172.21.0.14:36924: runtime error: invalid memory address or nil pointer dereference
> ```

Are you sure, that this is the only error message? When there is a panic, there should be a lot of lines in the error log

bastianjoel:
I cannot replicate this

jsangmeister:
> I cannot replicate this

Me neither

PR #847, opened 2024-02-14
Add motion restriction mode for `is_internal`

### LINKED ISSUE #840: Add motion state for internal recommendations
extracted from https://github.com/OpenSlides/openslides-autoupdate-service/issues/816

The requirements changed a little as opposed to what was stated in the old issue. The new situation is the following:

A motion state has the new boolean flag `is_internal` (renamed from `is_internal_recommendation` - I will change this in the backend). If a motion is in state with `is_internal == true`, its `recommendation_id` should be restricted to be only visible to users with `motion.can_manage`. This is independent of the recommendation!

@ostcar please reevaluate how feasible this is and, if possible, implement it.

bastianjoel:
resolves #840

### DIFF
--- a/internal/restrict/collection/motion.go
+++ b/internal/restrict/collection/motion.go
@@ -28,7 +28,7 @@ import (
 //
 // Mode D: Never published to any user.
 //
-// Mode E: If is_internal is set the user needs the permission motion.can_manage otherwise same as Mode C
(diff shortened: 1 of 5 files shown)


ostcar:
Looks perfect. Thank you

ostcar:
I thought this was a pull request. If you don't want it to be merged, then please open it as a draft.

PR #848, opened 2024-02-14
Rename internal recommendatio field

### LINKED ISSUE #840: Add motion state for internal recommendations

jsangmeister:
fixes #840

Actual restrictions still missing, will be done by @bastianjoel

### DIFF
--- a/internal/restrict/field_def.go
+++ b/internal/restrict/field_def.go
@@ -884,7 +884,7 @@ var restrictionModes = map[string]string{
 	"motion_state/css_class":                           "A",
 	"motion_state/first_state_of_workflow_id":          "A",
 	"motion_state/id":                                  "A",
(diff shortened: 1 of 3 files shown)


jsangmeister:
superseded by https://github.com/OpenSlides/openslides-autoupdate-service/pull/847/

PR #849, opened 2024-02-14
Yaml

ostcar:
This uses the new yaml lib. It seems to work.

@jsangmeister and @bastianjoel Is this the way to go? I don't see any harm in mergin it.


jsangmeister (approved):
Looks good from my amateur perspective if it works. @bastianjoel could you do the same for the search service?

bastianjoel (approved):

Issue #870, opened 2024-03-01
Prevent Error talking to autoupdate as Deli if not allowed to seeLoS


Elblinator:
**Current behaviour:**
If a participant is allowed to put themselves on the LoS but not allowed to see the LoS then opening some pages produces an "Error talking to autoupdate service"

**Reproduction:**
1. Participant A group settings for A:
2. A needs the right to put themselves on the LoS
3. A is not allowed to see the LoS
4. A is allowed to see the agenda and the front page
5. Login as A
6.  in a meeting: navigate to the start page
7. reload the page
8. navigate to the agenda
9. error talking to autoupdate occurs

**Logs from the autoupdate:**
```
autoupdate_1        | 2024/03/01 12:10:15 http: panic serving 172.22.0.14:42328: runtime error: invalid memory address or nil pointer dereference
autoupdate_1        | goroutine 2939 [running]:
autoupdate_1        | net/http.(*conn).serve.func1()
autoupdate_1        |   /usr/local/go/src/net/http/server.go:1898 +0xbe
autoupdate_1        | panic({0xabfe40?, 0x1146080?})
autoupdate_1        |   /usr/local/go/src/runtime/panic.go:770 +0x132
autoupdate_1        | github.com/OpenSlides/openslides-autoupdate-service/pkg/datastore/dsfetch.(*ValueInt).Lazy(...)
autoupdate_1        |   /root/pkg/datastore/dsfetch/fields_generated.go:195
autoupdate_1        | github.com/OpenSlides/openslides-autoupdate-service/internal/restrict/collection.Speaker.see.func1(0x2, {0xc00052a880, 0xa, 0x10})
autoupdate_1        |   /root/internal/restrict/collection/speaker.go:73 +0x533
autoupdate_1        | github.com/OpenSlides/openslides-autoupdate-service/internal/restrict/collection.eachMeeting({0xc96d70, 0xc0000ad860}, 0xc0007c26f0, {0xc95b00, 0x11c8e20}, {0xc00054e780, 0xa, 0xc000101008?}, 0xc0000febc8)
autoupdate_1        |   /root/internal/restrict/collection/collection.go:260 +0x3cf
autoupdate_1        | github.com/OpenSlides/openslides-autoupdate-service/internal/restrict/collection.Speaker.see({}, {0xc96d70?, 0xc0000ad860?}, 0xc0000fec28?, {0xc00054e780?, 0xa89220?, 0xc0000fec01?})
[...]

PR #872, opened 2024-03-02
Add Maybe-Type for non required fields

ostcar:
I was a bit shocked, that there is a panic in issue #870. So I tried to make the fsfetch-Type-System of the autoupdate-service more typesave.

The problem was on Lazy-Types (needed to fetch many values from the DB at the same time) and values, that are not required.

Not required values can be `null`, but go types can not. This means,  that the go value `0` could mean, that there is `zero` in the DB or there could be `0` in the Database.

This PR fixes this, by adding a new `Maybe`-Type. `Maybe` is a concept, that some programming langues use as a wrapper around a type, so this value can either be `null` or a concrete value. With go-generics, it is easy to build this kind of a type-wrapper for go.

With this change, it is impossible as a programmer, to forget to check for zero values, as I did for the speaker-restrictor.

This change should also fix the issue #870, but I don't want to close it, until I understand it better.

### DIFF
--- a/internal/autoupdate/autoupdate.go
+++ b/internal/autoupdate/autoupdate.go
@@ -210,7 +210,7 @@ func (a *Autoupdate) skipWorkpool(ctx context.Context, userID int) (bool, error)
 			return false, fmt.Errorf("getting groupIDs of user: %w", err)
 		}
 
-		adminGroups := make([]int, len(groupIDs))
+		adminGroups := make([]dsfetch.Maybe[int], len(groupIDs))
 		for i := 0; i < len(groupIDs); i++ {
 			ds.Group_AdminGroupForMeetingID(groupIDs[i]).Lazy(&adminGroups[i])
 		}
@@ -268,7 +268,7 @@ func (a *Autoupdate) CanSeeConnectionCount(ctx context.Context, userID int) (boo
 		groupIDs = append(groupIDs, groupPerMeetingIDs[i]...)
 	}
 
-	isAdminGroup := make([]int, len(groupIDs))
+	isAdminGroup := make([]dsfetch.Maybe[int], len(groupIDs))
 	for i, groupID := range groupIDs {
 		ds.Group_AdminGroupForMeetingID(groupID).Lazy(&isAdminGroup[i])
 	}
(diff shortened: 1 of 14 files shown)


internal/autoupdate/autoupdate.go
```diff
@@ -210,7 +210,7 @@ func (a *Autoupdate) skipWorkpool(ctx context.Context, userID int) (bool, error)
 			return false, fmt.Errorf("getting groupIDs of user: %w", err)
 		}
 
-		adminGroups := make([]int, len(groupIDs))
+		adminGroups := make([]dsfetch.Maybe[int], len(groupIDs))
```
ostcar, line 213: Wow, in line 222 is a big bug!!!

it has to be `for _, isAdmin`. If I understand the code correctly, then a user can skip the workpool, if he is in more then one group. Can that be true? In this case, the workpool feature would be totally broken?

@bastianjoel could you look at this code and tell me, if I am correct? If we fix this bug, it could have a big impact on the speed, that users get there data. If the current implementation was not a big problem, then maybe, we have to rethink the workpool-feature.

The variable `adminGroups` is a slice with the len of all groups of the user in a meeting. For example, if the user is in 5 groups, then `adminGroups` is a slice with len 5. After line 218, it contains the field `group/admin_group_for_meeting_id` of all the groups from the user. So for a user, that is not an admin, but is in 5 groups, it is `[0,0,0,0,0]`. The for-loop in line 222 does not go over the values, but over the index. So in this case, `isAdmin` is `0,1,2,3,4`.

bastianjoel: > it has to be for _, isAdmin. If I understand the code correctly, then a user can skip the workpool, if he is in more then one group. Can that be true? In this case, the workpool feature would be totally broken?

Yes, this seems to be the case. 

> @bastianjoel could you look at this code and tell me, if I am correct? If we fix this bug, it could have a big impact on the speed, that users get there data. If the current implementation was not a big problem, then maybe, we have to rethink the workpool-feature.

I think we should just fix this for now and maybe reevaluate if this has performance implications. Maybe this isn't that much of a problem anymore as we optimized the fields requested from the client and the overall serving time improved that way.

bastianjoel (approved):

internal/autoupdate/autoupdate.go
ostcar: I created a separate issue for this, since this is not part of this PR. I will create a fix and discuss with @peb-adr , that he has a look at the performance

PR #890, opened 2024-03-14
Change LoS projecion for interposed question

luisa-beerboom:
Necessary changes for OpenSlides/openslides-client#3410

That issue requires the current speaker to remain the same, even if he is paused for the sake of an interposed question and to have some method to recognize whether he and the waiting interposed questions are currently speaking (for that purpose I added `begin_time´ and `pause_time` to the return payload)


bastianjoel:
Resolves #894 now too.

ostcar (approved):

PR #907, opened 2024-04-05
Add yaml parser workaround for anchors

bastianjoel:
needed for https://github.com/OpenSlides/openslides-meta/pull/69


jsangmeister (approved):
tested

ostcar:
I am not a fan. I don't like, that the data it decoded, encoded again and decoded a second time.

I want to try something. If it doesn't work, we can do it this way

jsangmeister:
> I am not a fan. I don't like, that the data it decoded, encoded again and decoded a second time.

Yes, it's not optimal, but it's the best we could come up with. Currently, the parser is faced with two problems:
- There is no easy way to *not* parse the `_meta` key as a model, but as a separate attribute with a different interface
- The YAML lib does handle anchors/aliases with different indentations correctly when using custom unmarshalers

This approach solves both of these problems in one go (pun not intended). If you're solution solves those as well in a better way, I would be happy to accept it.

ostcar:
I was not able to find a better solution. So lets use it.

For the future, we could thing about changing the format of the models.yml so something like this:

meta: ...
config:...
other_fields: ...
collection:
  motion:
     ...
   agenda_item:
      ...

ostcar (approved):

bastianjoel:
We tried something similar but then stumbled over https://github.com/goccy/go-yaml/issues/431

bastianjoel:
The issue was marked as resolved. We might want to check if we can remove the workaround now.

PR #920, opened 2024-04-17
Fix restricter on vote delegation

ostcar:
This fixes the bug, that for some cases, only "undefined" is shown in the motion list for submitters.

There where two bugs. The core bug was a very strange case in the "required user to see"-list. I don't know, why I added it. It is not in the documentation and it is clearly wrong. It uses the field `MeetingUser_VoteDelegationsFromIDs` and uses the returned meeting_user_ids in the `Vote`-collection. Since the vote with this ID did not exist, an error was thrown.

The second bug was, that this error was silently ignored. I can not remember, why I ignored `DoesNotExist` errors at this place. There might be a reason. But If I don't ignore it, the first bug would have be found very easily.

This PR does not ignore this kind of errors anymore. This could mean, that there are more 500er from the autoupdate service. Please test this PR, before deploying it.


ostcar:
@peb-adr I ping you, so you are informed, if you are getting 500er in the autoupdate-service, after this is merged

rrenkert (approved):

PR #934, opened 2024-05-10
Remove history

ostcar:
This removes the ability of the autoupdate-service to access the history data. So an superadmin can not do this by accident.

The history information can still be accessed. But the new implementation fetches this data directly from postgres instead of calling the datastore-reader. With this change, the autoupdate-service will never call the datastore-reader. The environment variables to connect to the datastore-reader are not needed anymore.


ostcar:
I see, that the corresponding work in the Client was already done. Can I merge this?

bastianjoel:
Yes, this can be merged

bastianjoel (approved):

PR #935, opened 2024-05-15
Add title info to chyron

bastianjoel:
needed for https://github.com/OpenSlides/openslides-client/issues/3655

### DIFF
--- a/internal/projector/slide/list_of_speakers.go
+++ b/internal/projector/slide/list_of_speakers.go
@@ -148,6 +148,7 @@ func CurrentSpeakerChyron(store *projector.SlideStore) {
 		}
 
 		var shortName, structureLevel string
+		var titleInfo json.RawMessage
 		if losID > 0 {
 			shortName, structureLevel, err = getCurrentSpeakerData(ctx, fetch, losID, meetingID)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


ostcar (approved):

internal/projector/slide/list_of_speakers.go
```diff
@@ -156,16 +157,38 @@ func CurrentSpeakerChyron(store *projector.SlideStore) {
...
+
+			var losContentObject string
+			fetch.FetchIfExist(ctx, &losContentObject, "list_of_speakers/%d/content_object_id", losID)
+			if err := fetch.Err(); err != nil {
+				return nil, fmt.Errorf("getting los content object %d: %w", losID, err)
+			}
+
+			parts := strings.Split(losContentObject, "/")
+			if len(parts) != 2 {
+				return nil, fmt.Errorf("splitting ComtentObjectID: %w", err)
```
ostcar, line 169: The error message is wrong. `strings.Split` does not return an error. So the `err` variable will always be nil here. You have to change the error message to something like_ `fmt.Errorf("expeting ContentObjectID to have exacly one '/', got %d", len(parts) - 1)`

ostcar (changes requested):

ostcar (approved):

PR #966, opened 2024-07-12
Change restricter for user modeH

### LINKED ISSUE #965: Change user restriction mode H to allow committee admins
New description of H:
```
The fields can be seen if one of the following conditions is true:
- Any of the conditions in D or 
- Y has committee_management in any committee X is part of (i.e. also a committee manager or in any group or any meeting in the committee)
But the fields are never visible, if the request has a lower organization management level than the requested user.
```

ostcar:
Fixes #965

@luisa-beerboom I was not sure, what exactly ` Y has committee_management in any committee X is part of (i.e. also a committee manager or in any group or any meeting in the committee)`  means. So I changed it to `There exists a committee where Y has the CML can_manage and X is in committee/user_ids.` since this is the term we use in the other restriction modes.

If this is not correct, then I need a better explanation.

### DIFF
--- a/internal/restrict/collection/user.go
+++ b/internal/restrict/collection/user.go
@@ -54,8 +54,10 @@ import (
 //
 // Mode G: No one. Not even the superadmin.
 //
-// Mode H: Like D but the fields are not visible, if the request has a lower
-// organization management level then the requested user.
+// Mode H: The fields can be seen if one of the following conditions is true:
+//   - Any of the conditions in D or
+//   - There exists a committee where Y has the CML can_manage and X is in committee/user_ids
+//     But the fields are never visible, if the request has a lower organization management level than the requested user.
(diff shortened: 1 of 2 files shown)


luisa-beerboom:
> @luisa-beerboom I was not sure, what exactly ` Y has committee_management in any committee X is part of (i.e. also a committee manager or in any group or any meeting in the committee)` means. So I changed it to `There exists a committee where Y has the CML can_manage and X is in committee/user_ids.` since this is the term we use in the other restriction modes.
> 
> If this is not correct, then I need a better explanation.

That is exactly what I meant

luisa-beerboom (approved):

ostcar:
Should I merge this immediately or do we have to wait for something else?

bastianjoel (approved):

PR #971, opened 2024-07-24
Locked meeting

### LINKED ISSUE #969: Lock out meeting
For the new field `meeting/locked_from_inside` the restrictor has to be changed to lock out all admins. The superadmin can not be locked out, since he does not call the restrictor.

First step: Go though all restrictions mods to see, which have to be changed.

ostcar:
Fixes #969

### DIFF
--- a/internal/restrict/collection/meeting.go
+++ b/internal/restrict/collection/meeting.go
@@ -18,6 +18,8 @@ import (
 //	The user has the CML can_manage of any meeting and the meeting is a template meeting.
 //	The user has the OML can_manage_organization.
 //
+// If `meeting/locked_from_inside` is set, only users in the meeting can see it.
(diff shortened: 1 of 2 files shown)


ostcar:
@Elblinator To test this, you can run this with the current client (that does not implement the locked meeting yet). You should not see meeting data as an orga admin

Elblinator:
I still get all the data from the committee_detail:subscription as Orga Admin. 
But that is something I (naively) expected because no meeting is locked from the inside, and therefore the data should be returned(?).

I do not see faulty behaviour right now. If the code is fine we can merge this PR imo.

bastianjoel (approved):

ostcar:
closed in favor of #972

PR #972, opened 2024-07-24
Lock out user

ostcar:
This is not done yet. The meeting restrictor has to be changed. But this is in conflict with #971. I wait, until the other PR is merged

@Elblinator you can test this. The meeting may still appear, but there should not be any data in the meeting.


ostcar:
The user restrictor has to be changed, so that locked out users can not be seen in a meeting, only with the sensibleData permission

MSoeb:
Note about the group permission 'can_see_participants': We've discussed in the ABS-Team what should be the best practical solution for this permission. We've descided, that it would be good, if users with this permission can't see locked out users in the participant list, because there is no good or useful reason to see them. On the contrary, seeing users who are excluded from the meeting - for whatever reason - could lead to unnecessary questions, which we want to avoid.

For practical use, this means that users with the permission cannot see excluded users in the list of participants, but excluded users can still be submitters or supporters of applications, for example. It is only a matter of visibility in the list of participants.

ostcar:
I merged PR #971 and added the necessary changed. So if this gets merged, #969 is also fixed.

ostcar:
If you want to merge this, you have to check this branch out, update the meta subrepo and call go generate ./...

bastianjoel:
Is there a good reason why a user cannot see themselves whether they are locked out of a meeting? 

Without this information I cannot differentiate between content not getting loaded for whatever reason or the user being locked out. Therefore the infinite spinner described in https://github.com/OpenSlides/openslides-client/pull/3815#issuecomment-2291118528 is not fixable. 

@Elblinator @rrenkert

rrenkert:
If the user can see the meeting id and the 'locked out' attribute it would be the same as guessing the meeting id and enter ist in the url. So returning this data would be ok!

bastianjoel (approved):

PR #981, opened 2024-07-31
Server time

ostcar:
`/system/autoupdate/server_time` now returns something like:
```
{"server_time": 1722429116}
```

I also cleaned up some stuff with that PR, where go vet was complaining about


bastianjoel (approved):
LGTM ~~but there are some changes I am not sure were intended to be included here. ~~ Nevermind

PR #982, opened 2024-07-31
Anonymous only

ostcar:
Add new anonymous only mode

* Does not connect to the vote service
* Does not save in the redis metric
* Does not show values from the redis metric


m-schieder (approved):

internal/http/http.go
```diff
@@ -43,12 +43,10 @@ func Run(
...
-	if redisConnection != nil {
-		connectionCount[0] = newConnectionCount(ctx, redisConnection, saveIntercal, "connections_stream")
-		connectionCount[1] = newConnectionCount(ctx, redisConnection, saveIntercal, "connections_longpolling")
-		metric.Register(connectionCount[0].Metric)
-		metric.Register(connectionCount[1].Metric)
-	}
+	connectionCount[0] = newConnectionCount(ctx, redisConnection, saveIntercal, "connections_stream")
+	connectionCount[1] = newConnectionCount(ctx, redisConnection, saveIntercal, "connections_longpolling")
+	metric.Register(connectionCount[0].Metric)
+	metric.Register(connectionCount[1].Metric)
```
bastianjoel, line 49: I am not sure if this check should be removed.

ostcar: The check is moved down. Before this PR, this check was a hack for the tests. For the tests, I called the function with redisConnection == nil, so I had not to start a redis server.

Now, redisConnection is nil on an anonymous_only instance. In this case, the logic happens in [internal/http/connection_count.go:42](https://github.com/OpenSlides/openslides-autoupdate-service/pull/982/files#diff-9f83667bf40290fb5b85c44b1a642ed6c44d9cfd6dd3ec6701797c98a64ecf01R41)

So now, it is a valid behaviour to call the function with `nil`

bastianjoel (approved):

PR #1000, opened 2024-08-30
Public organization mediafiles

bastianjoel:
This might need so discussion as with this orga media files get restricted more restrictive when they are published in organization which seems a bit counter intuitive and is not yet documented in the wiki. 

Rest should be according to https://github.com/OpenSlides/OpenSlides/wiki/Mediafiles


bastianjoel:
Discussed with @emanuelschuetze that orga mediafiles should require can manage organization permissions.

internal/restrict/collection/organization.go
```diff
@@ -59,3 +63,12 @@ func (Organization) modeC(ctx context.Context, ds *dsfetch.Fetch, userIDs ...int
 
 	return nil, nil
 }
+
+func (Organization) modeE(ctx context.Context, ds *dsfetch.Fetch, ids ...int) ([]int, error) {
```
ostcar, line 67: Please add a test for the new modeE

internal/restrict/collection/organization.go
```diff
@@ -59,3 +63,12 @@ func (Organization) modeC(ctx context.Context, ds *dsfetch.Fetch, userIDs ...int
...
+	isAdmin, err := isMeetingAdmin(ctx, ds)
+	if err != nil || !isAdmin {
```
ostcar, line 69: Please do this in two steps. I understand, that it works, but it is not clear on the first view, that in the condition block, error can be nil, so the condition ins not only in error casses.

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
-
-		perms, err := perm.FromContext(ctx, ownerID)
-		if err != nil {
-			return nil, fmt.Errorf("getting perms for meeting %d: %w", ownerID, err)
-		}
-
-		if perms.IsAdmin() {
-			return ids, nil
-		}
+				managementLevel, err := perm.HasOrganizationManagementLevel(ctx, ds, requestUser, perm.OMLCanManageOrganization)
```
ostcar, line 80: In line 13 to line 27 is the documentation of the collection.  This probably also has to change.

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
+				if err != nil {
+					return nil, err
```
ostcar, line 82: Please add a new error string like `return nil, fmt.Errorf("....: %w", err)

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
-			return nil, fmt.Errorf("can see meeting %d: %w", ownerID, err)
+				if managementLevel {
+					return ids, nil
+				}
+			}
 		}
 
 		return eachCondition(ids, func(mediafileID int) (bool, error) {
-			logoOrFont, err := usedAsLogoOrFont(ctx, ds, mediafileID)
+			token, err := ds.Mediafile_Token(mediafileID).Value(ctx)
```
ostcar, line 92: Please add an error string in in line 94

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
 			if err != nil {
 				return false, err
 			}
 
-			if len(canSeeMeeting) == 1 && logoOrFont {
+			if token != "" {
```
ostcar, line 97: Are you sure, you don't have to check the token?

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
 				return true, nil
 			}
 
-			if perms.Has(perm.ProjectorCanSee) {
-				p7onIDs, err := ds.Mediafile_ProjectionIDs(mediafileID).Value(ctx)
+			published, err := ds.Mediafile_PublishedToMeetingsInOrganizationID(mediafileID).Value(ctx)
+			if err != nil {
+				return false, err
```
ostcar, line 103: Please add an error string

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
+			}
+
+			if val, isSet := published.Value(); isSet && val == 1 {
```
ostcar, line 106: I think, I would have written: `if val, _ := published.Value(); val == 1`. If isSet is false, then val is always 0. But maybe your solution is more clear.

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
+				isAdmin, err := isMeetingAdmin(ctx, ds)
 				if err != nil {
-					return false, fmt.Errorf("getting projection ids: %w", err)
+					return false, err
```
ostcar, line 109: Please add an error string

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
-					}
-
-					if !value.Null() {
-						return true, nil
-					}
+				if isAdmin {
+					return true, nil
 				}
+			} else if collection == "organization" {
+				return requestUser != 0, nil
```
ostcar, line 116: This is complicated to read. There are a lot of conditions here, that are in different lines.

It returns true here, if requestUser is not 0, and collection == "organization" and it is not published or it is published, but the value is not 1.

Could you restructure the code, so the different conditions get more clear? I think a switch statement could help

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
 			}
 
-			if perms.Has(perm.MediafileCanManage) {
-				return true, nil
+			meetingMediafileIDs, err := ds.Mediafile_MeetingMediafileIDs(mediafileID).Value(ctx)
+			if err != nil {
+				return false, err
```
ostcar, line 121: Please add an error string

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +77,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
 			}
 
-			if perms.Has(perm.MediafileCanSee) {
-				public, err := ds.Mediafile_IsPublic(mediafileID).Value(ctx)
+			if len(meetingMediafileIDs) > 0 {
+				canSeeMeetingMediafile, err := Collection(ctx, MeetingMediafile{}.Name()).Modes("A")(ctx, ds, meetingMediafileIDs...)
```
ostcar, line 125: With this call, Mediafile depends on MeetingMediaFile. So you have to make sure, that MeetingMediaFile is above Mediafile in restrict.go line 582. This makes sure, MeetingMediaFileIDs are called before Mediafile, so the caching works.

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,183 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/OpenSlides/openslides-autoupdate-service/internal/restrict/perm"
+	"github.com/OpenSlides/openslides-autoupdate-service/pkg/datastore/dsfetch"
+)
+
+// MeetingMediafile handels permissions for the collection mediafile.
```
ostcar, line 11: It seems, that this is a copy and paste of the Mediafile. Could you check this? For example, what is a MeetingMediafile, that belongs to an organization

internal/restrict/restrict.go
```diff
@@ -579,4 +579,5 @@ var collectionOrder = map[string]int{
 	"structure_level_list_of_speakers": 43,
 	"motion_working_group_speaker":     44,
 	"motion_editor":                    45,
+	"meeting_mediafile":                46,
```
ostcar, line 582: This probably has to be above mediafile (currently on 9). See above

ostcar (changes requested):
I could only review the code but not, if it is doing the correct thing.

Please first update the Docstring above Mediafile and MeetingMediafile so I can see, how it should behave. Then I can review it again.

internal/restrict/collection/mediafile.go
bastianjoel: In https://github.com/OpenSlides/OpenSlides/wiki/Mediafiles#organization-wide-mediafiles it is documented that a file should be public with any token.

internal/restrict/collection/mediafile.go
bastianjoel: The specifications changed here.

internal/restrict/collection/mediafile.go
bastianjoel: This got a bit simpler now with a specification change.

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/OpenSlides/openslides-autoupdate-service/internal/restrict/perm"
+	"github.com/OpenSlides/openslides-autoupdate-service/pkg/datastore/dsfetch"
+)
+
+// MeetingMediafile handels permissions for the collection meeting_mediafile.
+//
+// The user can see a meeting mediafile if any of:
+//
+//	The user is an admin of the meeting.
```
ostcar, line 15: I think this can be removed. An admin of a meeting automaticity has `mediafile.can_manage`, so this rule is redundant.

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+//	The user can see the meeting and used_as_logo_*_in_meeting_id or used_as_font_*_in_meeting_id is not empty.
+//	The user has projector.can_see and there exists a meeting_mediafile/projection_ids with projection/current_projector_id set.
```
ostcar, line 17: Could you change this sentence? It seems wrong. I think what you mean is:
```
The user has projector.can_see and one of the projections linked with meeting_mediafile/projection_ids has projection/current_projector_id set.
```

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (m MeetingMediafile) see(ctx context.Context, ds *dsfetch.Fetch, meetingMediafileIDs ...int) ([]int, error) {
+	return eachMeeting(ctx, ds, m, meetingMediafileIDs, func(meetingID int, ids []int) ([]int, error) {
+		perms, err := perm.FromContext(ctx, meetingID)
+		if err != nil {
+			return nil, fmt.Errorf("getting perms for meeting %d: %w", meetingID, err)
+		}
+
+		if perms.IsAdmin() {
```
ostcar, line 57: This can be changed with
```
if perms.Has(perm.MediafileCanManage) {
    return ids, nil
}
```

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+						return false, fmt.Errorf("getting current projector id: %w", err)
+					}
+
+					if !value.Null() {
+						return true, nil
+					}
+				}
+			}
+
+			if perms.Has(perm.MediafileCanManage) {
```
ostcar, line 94: This is not needed, if you change the `isAdmin` before the loop

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+		perms, err := perm.FromContext(ctx, meetingID)
+		if err != nil {
+			return nil, fmt.Errorf("getting perms for meeting %d: %w", meetingID, err)
+		}
+
+		if perms.IsAdmin() {
+			return ids, nil
+		}
+
+		canSeeMeeting, err := Collection(ctx, Meeting{}.Name()).Modes("B")(ctx, ds, meetingID)
```
ostcar, line 61: You can use this as an early escape. In the user can not see a meeting, then non of the rules above match.

So you can add a `if !canSeeMeeting { return nil, nil}`

This should be at the top of the `eachMeeting`

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+			return nil, fmt.Errorf("can see meeting %d: %w", meetingID, err)
+		}
+
+		return eachCondition(ids, func(meetingMediafileID int) (bool, error) {
+			logoOrFont, err := usedAsLogoOrFont(ctx, ds, meetingMediafileID)
+			if err != nil {
+				return false, err
+			}
+
+			if len(canSeeMeeting) == 1 && logoOrFont {
```
ostcar, line 72: When you add the early escape of `canSeeMeeting` above, then you don't need the condition here.

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+//
+// The user can see a meeting mediafile if any of:
+//
+//	The user is an admin of the meeting.
+//	The user can see the meeting and used_as_logo_*_in_meeting_id or used_as_font_*_in_meeting_id is not empty.
+//	The user has projector.can_see and there exists a meeting_mediafile/projection_ids with projection/current_projector_id set.
+//	The user has mediafile.can_manage.
+//	The user has mediafile.can_see and either:
+//	    meeting_mediafile/is_public is true, or
+//	    The user has groups in common with meeting_mediafile/inherited_access_group_ids.
```
ostcar, line 21: This is not a full sentence. The `The` at the beginning has to be lowercase.

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -0,0 +1,181 @@
...
+					return true, nil
+				}
+
+				inheritedGroups, err := ds.MeetingMediafile_InheritedAccessGroupIDs(meetingMediafileID).Value(ctx)
+				if err != nil {
+					return false, fmt.Errorf("getting inheritedGroups: %w", err)
+				}
+
+				for _, id := range inheritedGroups {
+					if perms.InGroup(id) {
```
ostcar, line 125: I don't like this loop. `eachCondition` loops throw all mediafile_ids. then there is the loop over inheritedGroups and inside of `perms.InGroup` there is a loop over the groups of the user. So this is O(n^3). Maybe change [pers.InGroup](https://github.com/OpenSlides/openslides-autoupdate-service/blob/fd44caacbd50854f4fb7f9ac7c8c4be874cba4b7/internal/restrict/perm/perm.go#L171-L186) to:
```go
func (p *Permission) InGroup(groupIDs ...int) bool {
	if p == nil {
		return false
	}

	if p.admin {
		return true
	}

        for_, gid := range groupIDs {
	    for _, id := range p.groupIDs {
		    if id == gid {
			return true
		    }
             }
	}
	return false
}
```
Then it is still O(N^3) but at least the two conditions only are checked once per mediafile*inheritedGroups (which still is too many times, but at least...)

internal/restrict/collection/mediafile.go
```diff
@@ -12,17 +12,12 @@ import (
...
-// The user can see a mediafile of a meeting if any of:
-//
-//	The user is an admin of the meeting.
-//	The user can see the meeting and used_as_logo_*_in_meeting_id or used_as_font_*_in_meeting_id is not empty.
-//	The user has projector.can_see and there exists a mediafile/projection_ids with projection/current_projector_id set.
-//	The user has mediafile.can_manage.
-//	The user has mediafile.can_see and either:
-//	    mediafile/is_public is true, or
-//	    The user has groups in common with meeting/inherited_access_group_ids.
+//	The meeting mediafile belongs to the organization and the user has organization management permissions
```
ostcar, line 17: Would the statement be true, that with organization management permissions, you can see all `mediafiles`, even when they belong to a meeting? If this is true, the code could be more optimized.

But I am not sure, since the organization-manager could be looked out of the meeting.

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +72,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
-
-		perms, err := perm.FromContext(ctx, ownerID)
-		if err != nil {
-			return nil, fmt.Errorf("getting perms for meeting %d: %w", ownerID, err)
-		}
-
-		if perms.IsAdmin() {
-			return ids, nil
-		}
+				managementLevel, err := perm.HasOrganizationManagementLevel(ctx, ds, requestUser, perm.OMLCanManageOrganization)
```
ostcar, line 75: Please move this before the `eachContentObjectCollection`. The variable `managementLevel` is the same for each. So it can be calculated only once.

Then you canchange line 73 from `if collection == "organization" {` to `if collection == "organization" && managementLevel {` (you could also rename `managementLevel` to `hasManagementLevel`)

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +72,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
 
-			if perms.Has(perm.ProjectorCanSee) {
-				p7onIDs, err := ds.Mediafile_ProjectionIDs(mediafileID).Value(ctx)
+			published, err := ds.Mediafile_PublishedToMeetingsInOrganizationID(mediafileID).Value(ctx)
+			if err != nil {
+				return false, fmt.Errorf("getting published to meetings in organization: %w", err)
+			}
+
+			if val, _ := published.Value(); val == 1 {
+				isAdmin, err := isMeetingAdmin(ctx, ds)
```
ostcar, line 102: Please move this check at the top of the function. `isMeetingAdmin` should only be called once, and not once per mediafile.

Then you can also change the condition in line 101 from `val == 1 && isAdmin`

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +72,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
 			}
 
-			if perms.Has(perm.ProjectorCanSee) {
-				p7onIDs, err := ds.Mediafile_ProjectionIDs(mediafileID).Value(ctx)
+			published, err := ds.Mediafile_PublishedToMeetingsInOrganizationID(mediafileID).Value(ctx)
+			if err != nil {
+				return false, fmt.Errorf("getting published to meetings in organization: %w", err)
+			}
+
+			if val, _ := published.Value(); val == 1 {
```
ostcar, line 101: I think this reads strange. Maybe `if !published.Null()` is nicer?

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +72,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
-						return false, fmt.Errorf("getting current projector id: %w", err)
-					}
-
-					if !value.Null() {
-						return true, nil
-					}
+				if isAdmin {
+					return true, nil
 				}
+			} else if collection == "organization" {
```
ostcar, line 110: I still think, that an explicit

`if collection == "organization" && (!published.Null() || isAdmin) {` 

is easier to read

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +72,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
-			if perms.Has(perm.MediafileCanManage) {
-				return true, nil
+			meetingMediafileIDs, err := ds.Mediafile_MeetingMediafileIDs(mediafileID).Value(ctx)
+			if err != nil {
+				return false, fmt.Errorf("getting meeting mediafile ids: %w", err)
 			}
 
-			if perms.Has(perm.MediafileCanSee) {
-				public, err := ds.Mediafile_IsPublic(mediafileID).Value(ctx)
+			if len(meetingMediafileIDs) > 0 {
```
ostcar, line 119: I am not sure if this condition is needed. I think it should work without. A call to `Collection(ctx, MeetingMediafile{}.Name()).Modes("A")(ctx, ds, meetingMediafileIDs...)` with an empty list of meetingMediafileIDs should return an empty list of ids, so `if len(canSeeMeetingMediafile) >= 1 {` is still correct.

internal/restrict/collection/mediafile.go
```diff
@@ -77,137 +72,103 @@ func (m Mediafile) see(ctx context.Context, ds *dsfetch.Fetch, mediafileIDs ...i
...
-		UsedAsLogoPdfBallotPaperInMeetingID    dsfetch.Maybe[int]
-		UsedAsFontRegularInMeetingID           dsfetch.Maybe[int]
-		UsedAsFontItalicInMeetingID            dsfetch.Maybe[int]
-		UsedAsFontBoldInMeetingID              dsfetch.Maybe[int]
-		UsedAsFontBoldItalicInMeetingID        dsfetch.Maybe[int]
-		UsedAsFontMonospaceInMeetingID         dsfetch.Maybe[int]
-		UsedAsFontChyronSpeakerNameInMeetingID dsfetch.Maybe[int]
-		UsedAsFontProjectorH1InMeetingID       dsfetch.Maybe[int]
-		UsedAsFontProjectorH2InMeetingID       dsfetch.Maybe[int]
+func isMeetingAdmin(ctx context.Context, ds *dsfetch.Fetch) (bool, error) {
```
ostcar, line 135: The name is unclear. When I first read it, I though "which meeting?". Could you change it to something like `isAdminInAnyMeeting`? Or is this too long for an unexported function?

If you keep the name, could you at leat add a short doc-comment, that explains it?

internal/restrict/restrict.go
```diff
@@ -542,41 +542,42 @@ var collectionOrder = map[string]int{
...
-	"speaker":                          37,
-	"user":                             38,
-	"meeting_user":                     39,
-	"action_worker":                    40,
-	"import_preview":                   41,
-	"structure_level":                  42,
-	"structure_level_list_of_speakers": 43,
-	"motion_working_group_speaker":     44,
+	"meeting_mediafile":                9,
+	"mediafile":                        10,
```
ostcar, line 546: I am sorry. I should have used another system then a map[string]int. It is hard to change and not performant.
I don't know, why I did not use just a []string. But this is something for another PR :)

ostcar (changes requested):

internal/restrict/collection/meeting_mediafile.go
```diff
@@ -105,6 +105,17 @@ func (m MeetingMediafile) see(ctx context.Context, ds *dsfetch.Fetch, meetingMed
 					return true, nil
 				}
 
+				accessGroups, err := ds.MeetingMediafile_AccessGroupIDs(meetingMediafileID).Value(ctx)
+				if err != nil {
+					return false, fmt.Errorf("getting accessGroups: %w", err)
+				}
+
+				for _, id := range accessGroups {
```
ostcar, line 113: If you change `perms.InGroup` as proposed in another comment, then you can use `perms.InGroup(accessGroups...)` here as well

ostcar (changes requested):

internal/restrict/restrict.go
bastianjoel: Its just `qa<Ctrl>+aqj35@a` in vim, so no big deal ;)

internal/restrict/collection/meeting_mediafile.go
bastianjoel: I misread the documentation here. Access groups are not relevant.

internal/restrict/collection/meeting_mediafile.go
bastianjoel: `MediafileCanManage` will not permit a user to access all mediafiles anymore.

internal/restrict/collection/meeting_mediafile.go
bastianjoel: This is kept in favor of completely removing the `perm.MediafileCanManage` check.

internal/restrict/collection/meeting_mediafile.go
bastianjoel: Also changed in some other places

internal/restrict/collection/mediafile.go
bastianjoel: I am not exactly sure but I would guess no.

ostcar (approved):
Thank you

PR #1013, opened 2024-09-10
[Cherrt-Pick] Public organization mediafiles (#1000)

peb-adr:
Manual cherry-pick of #1000.
The pick-to-staging action fails, if the meta PR was not cherry-picked and the resulting commit checked out in the service PR first.


peb-adr:
@bastianjoel can you please resolve the remaining conflict in `pkg/datastore/dskey/gen_collection_fields.go`.

After this should be good to merge.

ostcar:
@peb-adr `pkg/datastore/dskey/gen_collection_fields.go` is a generated file. The easiest way to fix it is by just calling `go generate ./...`

PR #1018, opened 2024-09-11
Allow meeting user restriction mode E for related user

bastianjoel:
In #972 there was a comment missed which requested to allow restriction mode to be accessible by the requesting user themselves. 

https://github.com/OpenSlides/openslides-autoupdate-service/pull/972#issuecomment-2304265275


ostcar (approved):

reiterl:
It doesn't solve the issues.

reiterl:
A user with one meeting, which they is locked out, tries to log in.
=> In the operator:subscription the autoupdate service delivers one meeting_user object, but this doesn't include the 'meeting_id'. So the meeting_user object cannot be used in the code (e.g. to get the locked_out boolean).

reiterl:
@bastianjoel please look at it.

bastianjoel:
I think the best way to solve this is to add a new restriction mode to meetings `E`, for the field `meeting_user_ids` which is the same as `B` but allows locked out users. 

On the other hand I think we don't necessarily need the meeting id in the client.

bastianjoel:
I solved this in the client.

ostcar:
@bastianjoel what is the status of this? Can this me merged or are you waiting for something?

bastianjoel:
I am currently waiting if https://github.com/OpenSlides/openslides-client/pull/3815 needs additional requirements.

PR #1026, opened 2024-09-21
New rule for derived motions

ostcar:


ostcar:
TODO: change motion_state that anybody, that can see a motion in motion_state/motion_id can also see it. It is probably necessary to also show a state if the user has motion.can_see, if there is no motion in the state

ostcar:
Everything is done. The motion_state is fixed. It can be seen, if you have `motion.can_see` or can see a linked motion.

Other changes where not necessary. Specially there was no need to change `motion_submitter`, since it already is implemented in the way, that a user can see it, if he can see the motion

bastianjoel (approved):

PR #1048, opened 2024-10-16
Remove motion statute paragraf

ostcar:


ostcar:
@bastianjoel @Elblinator The problem was the projector. I have no idea how the projector works and the tests are the horror. I removed random stuff until the tests did not fail anymore.

Fu**, I forgot that I enabled automerge

ostcar:
Could you test, that the projector still works?

PR #1050, opened 2024-10-22
Add back motion text to projection

Elblinator:
The motion text was accidentally removed from the projection
This is just putting the earlier code back


ostcar:
Thank you!

PR #1053, opened 2024-10-25
Change restrictor for superadmin and orga admin

ostcar:
- The super admin has to go throw the restrictor
- The orga admin and superadmin are automaticly admin in every meeting, giving them all permissions in all meetings

This lets the orga admin inside every meeting.

I solved this by making the orga admin a admin in every meeting that is not locked from inside (same for super admin).

Is this what you want?


Elblinator:
The corresponding client issue is created https://github.com/OpenSlides/openslides-client/issues/4380

Elblinator:
With this PR the superadmin cannot see a meeting which is locked from inside. The superadmin should still be able to see the meeting (title, language, event location, meeting date, meeting_admins, tags, external ID, the jitsi settings -domain, room name, password-, the meeting_users, amount of motions from the meeting, which committee it belongs to)

@emanuelschuetze should the orgaadmin now also be able to see meetings they are not a part of if these meetings are locked from inside?

ostcar:
What collection mode has to change for that? Which fields?

Elblinator:
collection meeting.go restriction mode B

Elblinator:
The orga admin should get the same information (The orgaadmin should see locker_from_inside meetings)

ostcar:
Meeting/modeB are nearly all fields. If the superadmin can see them, then he is not locked out.

You asked for the following fields:

* title -> There is not such field. There is `meeting/name` but this is in mode A an can be seen by everyone.
* language -> This field is in mode A and can be seen by everybody.
* event location -> There is no such field. There is `meeting/location`. If this should be visible for orga- and superadmins, even when the meeting is locked, we would need a new restriction mode.
*  meeting date -> There is no such field. There is `meeting/start_time` and `meeting/end_time`. This fields would also be moved to the new restriction mode.
* meeting_admins -> There is no such field. There is `meeting/admin_group_id`. This one is really complicated. It would not be enough to put this field in its own restriction mode, but it would also be necessary change the restrictor of group and meeting_user just for this special case.
* tags -> There is no such field. There is a field `tag_ids`. This field would be put in the new restriction mode and the restrictor for tag would need to be changed, that the superadmin can see all tags.
*  external ID -> The field `meeting/external_id` is in mode A
* the jitsi settings -domain, room name, password -> The fields `meeting/jitsi_domain`, `meeting/jitsi_room_name` and `meeting/jitsi_room_password` are in modeE. This mode can be seen by superadmins but not from orga admins, if the meeting is locked. If the orga admin should also see them, then modeE is not needed anymore.
* the meeting_users -> Are you sure? I don't understand, why this is necessary to see a list of meetings. If you really want this, then the meeting_user restrictor has to be changed, so that the superadmin and orga_admin can allways see all meeting_users.
*  amount of motions from the meeting -> There is no such field. If you want this, we would need such a field that is generated by the backend.
*  which committee it belongs to -> The field `meeting/committee_id` has to be moved to the new restriction mode.

Please tell me, if this is what you want.
* The orga admin should always see the jitsy settings
* The super and orga admin should see meeting_users from locked out meetings
* If it is really necessary, that the superadmin and orgaadmin can see the admins of a locked meeting

Elblinator:
The orga admin should always see the jitsy settings
 - no

The super and orga admin should see meeting_users from locked out meetings
- The superadmin/orgaadmin should only be able to see these fields from the meetings_user:  user_id, meeting_user, meeting_id


If it is really necessary, that the superadmin and orgaadmin can see the admins of a locked meeting
- yes

additional answers:
- Everything which is in Mode A is fine and can stay there
- which committee it belongs to -> The field meeting/committee_id has to be moved to the new restriction mode. yes please

ostcar:
This should implement everything.

For simplification I implemented, that orga managers (and supersuper) can see all groups of every meeting. Even if the meeting is locked. It would only be necessary for admin groups, but this distinction seemed strange and overcomplicated.

I did not understand, what you mean with `The superadmin/orgaadmin should only be able to see these fields from the meetings_user: user_id, meeting_user, meeting_id`  I changed the restriction mode for `user_id` and `meeting_id`. But I don't understand what you mean with `meeting_user`

PR #1123, opened 2025-01-30
Add single nominal vote poll slide

luisa-beerboom:
See OpenSlides/openslides-client#4563

This doesn't work yet. I don't know why.
Also the actual vote data still needs to be added.

What would be required:
The projection data needs to contain who was eligible and who voted (already covered by `entitled_users_at_stop`), how they voted (not covered by `entitled_users_at_stop`), their name information (I tried to add it into `entitled_users_at_stop` but couldn't get it to work).
Name information should probably not be pre-rendered into the form that should be displayed in the client, since that would not be accurately sortable (because the `title` attribute is not supposed to factor into the sorting). Alternatively one could of course provide the specific user vote data nicely formed into its own data structure and provide that data structure in a sorted manner. 

There's also the matter of the two new meeting settings `motion_poll_projection_name_order_first` and `motion_poll_projection_max_columns`:
The former is important for the name formatting and the sorting order and would need to be considered in in case the autoupdate pre-formats, the latter is only important for the display in the client. I don't think delivering these settings as part of the data is explicitly necessary, since they should be visible to anyone, who can see the projections, anyway.


bastianjoel:
Do not merge yet.

internal/projector/slide/poll.go
```diff
@@ -223,6 +259,48 @@ func getOptions(ctx context.Context, fetch *datastore.Fetcher, store *projector.
 	return options, nil
 }
 
+func getPollUser(ctx context.Context, fetch *datastore.Fetcher, userID int) (user *pollUserRepr, err error) {
+	data := fetch.Object(ctx, fmt.Sprintf("user/%d", userID), "id", "title", "first_name", "last_name")
```
ostcar, line 263: I am not sure. do you need a error check after this line?
```go
if err := fetch.Err(); err != nil {
		return nil, err
	}
```

internal/projector/slide/poll.go
```diff
@@ -223,6 +259,48 @@ func getOptions(ctx context.Context, fetch *datastore.Fetcher, store *projector.
...
+	user, err = pollUserFromMap(data)
+	if err != nil {
+		return nil, fmt.Errorf("get user data: %w", err)
+	}
+
+	return user, nil
+}
+
+func getVotes(ctx context.Context, fetch *datastore.Fetcher, optionID int) (votes []*pollVoteRepr, err error) {
+	voteIDs := datastore.Ints(ctx, fetch.FetchIfExist, "option/%d/vote_ids", optionID)
```
ostcar, line 273: Also here. I don't remember the API anymore. Is an error check necessary?

internal/projector/slide/poll.go
```diff
@@ -223,6 +259,48 @@ func getOptions(ctx context.Context, fetch *datastore.Fetcher, store *projector.
...
+
+	fetchFields := []string{
+		"id",
+		"user_id",
+		"value",
+	}
+
+	for _, voteID := range voteIDs {
+		data := fetch.Object(ctx, fmt.Sprintf("vote/%d", voteID), fetchFields...)
```
ostcar, line 282: Also here? I am not sure, if the error check in line 297 is enough. Maybe `data` could be `nil` if an error happens. In this case, I am not sure, if `pollVoteFromMap` can handle a nil value

internal/projector/slide/poll.go
```diff
@@ -245,3 +323,80 @@ func getTitleInfoFromContentObject(ctx context.Context, fetch *datastore.Fetcher
...
+				return fmt.Errorf("encoding entitled users interpretation")
+			}
+
+			entry["user"] = user
+			newUserData = append(newUserData, entry)
+		}
+
+		var pollUserDataJSON, err = json.Marshal(newUserData)
+		if err != nil {
+			return fmt.Errorf("encoding entitled users interpretation")
```
ostcar, line 394: Maybe add a `: %w` to the error message.

ostcar (approved):
Only hints. I don't know if the suggested changes are necessary.

PR #1132, opened 2025-02-03
Create a new restriciton mode for meeting, that lets anyone see its f…

ostcar:
TODOs:
- [x] Think about locked meeting
- [x] Change the restriction mods in meta


ostcar:
Should be ready. @Elblinator could you also have a look? I think we talked about the last change in this part.

Elblinator (approved):
As far as I can see the correct thing is displayed

PR #1137, opened 2025-02-05
Issue 1127

ostcar:


internal/restrict/collection/motion_change_recommendation_test.go
```diff
@@ -97,10 +120,36 @@ func TestMotionChangeRecommendationModeA(t *testing.T) {
...
+                meeting_id: 30
+                state_id: 10
+                all_derived_motion_ids: [10]
+            10:
+                meeting_id: 31
+                state_id: 10
+
+        motion_change_recommendation/1:
+            meeting_id: 30
+            motion_id: 10
```
bastianjoel, line 148:
```suggestion
            motion_id: 1
```

bastianjoel (changes requested):

internal/restrict/collection/motion_change_recommendation_test.go
ostcar: Thank you. Good catch. I used `3` instead of `1`. I try to use unique ids for all collections. I had bugs in old tests, where I accidentally fetched the wrong collection, but the test did pass, since the same id was used (`motion/1/id`, instead of `motion_change_recommendation/1/id`)

bastianjoel (approved):

PR #1163, opened 2025-03-10
Hide origin from closed meeting

bastianjoel:
https://github.com/OpenSlides/openslides-client/pull/4144#pullrequestreview-2643593366
> Clsoed meetings: If a closed meeting is a source meeting of a motion, the slide controller should not available, no tab should be shown and the meeting name should have a small closed lock icon right to it.


internal/restrict/collection/motion.go
```diff
@@ -87,6 +87,15 @@ func (m Motion) see(ctx context.Context, ds *dsfetch.Fetch, motionIDs ...int) ([
 					return true, nil
 				}
 
+				lockedMeeting, err := ds.Meeting_LockedFromInside(meetingID).Value(ctx)
```
ostcar, line 90: Please also update the documentation in the lines 13ff. For a review, it is easier for me to read in the docs, what the result should be, instead guessing from the code.

When I make updates to the restrictor, I first update the documentation, then the tests and the code last. This is a workflow that worked very well for me

ostcar, line 90: This check only needs the meeting-id. At least the DB-Request (line 90) should be outside of `eachCondition`. 

The most performance would be, if the permission check from line 86 and the locked meeting check would also be outside of `eachCondition`. But this is not so important and could make the code match more complicated.

internal/restrict/collection/motion_change_recommendation_test.go
```diff
@@ -148,6 +150,8 @@ func TestMotionChangeRecommendationModeA(t *testing.T) {
             motion_id: 3
 
         motion_state/10/id: 10
+
+				meeting/30/locked_from_inside: false
```
ostcar, line 154: The indentation seems wrong

ostcar (changes requested):

ostcar (approved):

PR #1193, opened 2025-05-07
Return `true` on `isAdminInAnyMeeting` as orgaadmin

### LINKED ISSUE #1192: Field "organization/published_mediafile_ids" restriction excludes superadmin and orga admins
In https://github.com/OpenSlides/openslides-client/issues/4931 was a bug reported about mediafiles.
It seems to me, that the field organization/published_mediafile_ids is restricted for superadmins and organization admins.
It has restrict mode E. "At least meeting admin in one meeting"
Is it possible to allow superadmins and orgaadmins to see this field?

bastianjoel:
resolves #1192


ostcar (changes requested):
Please add a description of the case in the docstring at the top of the file (line 17ff).

If you are saying, that "the user is admin in a meeting" also measn orga admins or superadmins, then this should be included in the description. For example:

The user is admin in a meeting or is orga admin and ....

The rest is perfect.

MSoeb (approved):

PR #1205, opened 2025-05-22
Dockerfile Rework

Janmtbehrens:
Reworked Dockerfile & Makefiles, so that there is one singular Dockerfile that supports development, testing and production images.
Also adding features like healthchecks and use of restricted app users, as well as fixing some bugs

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -28,7 +28,7 @@ jobs:
 
   readme-validation:
     # This Job confirms the validity and functionality of all functions presented in the README.md file
-    name: Readme Validation
+    name: Readme
     runs-on: ubuntu-latest
     services:
       redis:


.github/workflows/continuous-integration.yml
```diff
@@ -1,10 +1,47 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run Test
+        uses: OpenSlides/OpenSlides/dev/actions/build-and-test-service@main
+        with:
+          service: autoupdate
+
+  readme-validation:
```
ostcar, line 41: Is anything beneath needed? I created the workflows/readme.yml to test commands in the readme. But the readme needs to be cleaned up. I can do this after this PR is merged. Afterwards, I will cleanup this part to be the same as in the readme. Is this ok for your concept?

Dockerfile
```diff
@@ -1,46 +1,65 @@
+ARG CONTEXT=prod
```
ostcar, line 1: This line has no effect. It is in the global scope and gets reset at `FROM`. See https://docs.docker.com/build/building/variables/#scoping

Dockerfile
```diff
@@ -1,46 +1,65 @@
...
+
 FROM golang:1.24.4-alpine as base
-WORKDIR /root/openslides-autoupdate-service
 
-RUN apk add git
+## Setup
+ARG CONTEXT
+
+WORKDIR /app/openslides-autoupdate-service
+ENV APP_CONTEXT=${CONTEXT}
```
ostcar, line 9: Why is this necessary? The autoupdate-service does not read this environment variable

Dockerfile
```diff
@@ -1,46 +1,65 @@
...
+LABEL org.opencontainers.image.description="The Autoupdate Service is a http endpoint where the clients can connect to get the current data and also updates."
+LABEL org.opencontainers.image.licenses="MIT"
+LABEL org.opencontainers.image.source="https://github.com/OpenSlides/openslides-autoupdate-service"
 
+EXPOSE 9012
 
-# Test build.
-FROM base as testing
+## Healthcheck
+HEALTHCHECK CMD ["/app/openslides-autoupdate-service/openslides-autoupdate-service", "health"]
```
ostcar, line 29: The `LABEL`, `EXPOSE` and `HEALTHCHECK` defintions are confusing. In this current implementation, they are inherited to `dev` and `tests` but not to `prod`. `tests` does not need them, so the definition here is actually only used in `dev`.

Please move them to `dev` to make be more clear

Dockerfile
```diff
@@ -1,46 +1,65 @@
...
+# Production Image
 
+FROM base as builder
+RUN go build
+
+FROM scratch as prod
 
-# Productive build
-FROM scratch
+WORKDIR /
```
ostcar, line 52: This line has no effect and should be removed.

The prod image is `FROM scratch`. It does not inherit any values from other images. The default value for `WORKDIR` is `/`

Dockerfile
```diff
@@ -1,46 +1,65 @@
...
+ENV APP_CONTEXT=prod
```
ostcar, line 53: Please remove this line. The autoupdate service never reads this variable

Makefile
```diff
@@ -1,9 +1,16 @@
+SERVICE=autoupdate
+
 build-dev:
-	docker build . --target development --tag openslides-autoupdate-dev
+	bash ../dev/scripts/makefile/build-service.sh $(SERVICE) dev
```
ostcar, line 4: I really don't like this. The autoupdate-service is its own repo. This line expects to be called from the main repo. But if this line can only be called from the context of the main repo, this file should only exist in the main repo.

Please change the Makefile, so it also works, if only the autoupdate-service is checked out

```bash
mkdir only-autoupdate
cd only-autoupdate
git clone https://github.com/OpenSlides/openslides-autoupdate-service.git
cd openslides-autoupdate-service
make build-dev
```

The same for every other command in this file

dev/run-tests.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Executes all tests. Should errors occur, CATCH will be set to 1, causing an erroneous exit code.
```
ostcar, line 3: This comment is wrong. This file does not call all tests but only the tests, that does not need postgres.

Please make sure, that this script should not be used, if all tests should run

ostcar (changes requested):

Dockerfile
Janmtbehrens: That is not entirely true, it sets the default value for CONTEXT to be prod, so that the default target context is the production image. It will get reset with the next FROM, but because of the Line 6 'ARG CONTEXT' this global variable gets 'consumed' and is available for the build stage.
This was implemented with the intention that multiple different build stages can point to the same default value defined at the start of the Dockerfile instead of having multiple different default value declarations in the Dockerfile for each build stage. It may be unnecessarily verbose and is not specifically necessary for autoupdate, but I still do it like so here to uphold the same Dockerfile structure across all Services

dev/run-tests.sh
Janmtbehrens: This file is intended to run all tests. Can you please tell me which test commands are missing so that I can add them?

Janmtbehrens: Using a docker in docker setup, run-tests now executes dockertest tests as well. That should cover all available tests unless I missed something?

Makefile
Janmtbehrens: I have reverted the reference to the main repository build-service.sh script to a direct build command

ostcar (approved):
I don't like everything about this, but this seems like a good compromise.

PR #1229, opened 2025-07-12
Add restricter for new history collections

ostcar:
Also remove old history code.


ostcar:
Ok. Everything is done here. The history permissions are implemented the same as in the old system.

ostcar:
I needed to update the openslides-go repo. Could you try this again?

PR #1231, opened 2025-07-15
Implement live voting single vote slide

bastianjoel:


internal/projector/slide/poll.go
```diff
@@ -184,6 +188,12 @@ func pollSlideDataFunction(ctx context.Context, fetch *datastore.Fetcher, p7on *
 			"global_option_id",
 		}...)
 	}
+
+	liveVotingEnabled := datastore.Bool(ctx, fetch.FetchIfExist, "%s/%s", p7on.ContentObjectID, "live_voting_enabled")
```
ostcar, line 192: I am not sure, if you should depend on this settings. There are many conditions, that live-voting works (it has to be a named poll in the fast-backend). Maybe it is better to analyze the values in `live_votes`. If they have keys, it works, if not, it don't.

internal/projector/slide/poll.go
```diff
@@ -405,6 +415,59 @@ func PollSingleVotes(ctx context.Context, store *projector.SlideStore, fetch *da
...
+			User           *pollUserRepr `json:"user_data"`
+			Vote           *int          `json:"vote_id,omitempty"`
+			StructureLevel *int          `json:"structure_level_id,omitempty"`
+		}
+
+		liveVotingData := map[int]liveVotingEntry{}
+
+		var pollLiveVoteData map[int]*int
+		if err := json.Unmarshal(*poll.LiveVotes, &pollLiveVoteData); err != nil {
+			return fmt.Errorf("reading live vote data")
```
ostcar, line 429: a `: %w", err)` is missing

internal/projector/slide/poll.go
```diff
@@ -405,6 +415,59 @@ func PollSingleVotes(ctx context.Context, store *projector.SlideStore, fetch *da
...
+		}
+
+		structureLevels := map[int]string{}
+		for userID, voteID := range pollLiveVoteData {
+			user, err := getPollUser(ctx, fetch, userID)
+			if err != nil {
+				return fmt.Errorf("encoding live votes interpretation")
```
ostcar, line 436: Here the same.

ostcar (commented):
If you use [this branch of openslides-go](https://github.com/OpenSlides/openslides-go/pull/64) we could close this pr: https://github.com/OpenSlides/openslides-autoupdate-service/pull/1228/files

internal/projector/slide/poll.go
bastianjoel: IMHO the backend should just handle that this value is set correctly.

ostcar (approved):
I have not tested it, but looks good

PR #1236, opened 2025-07-28
Makefile Rework

Janmtbehrens:


Dockerfile
```diff
@@ -64,7 +76,10 @@ LABEL org.opencontainers.image.source="https://github.com/OpenSlides/openslides-
 
 COPY --from=builder /app/openslides-autoupdate-service/openslides-autoupdate-service /
 
+COPY --from=builder /app/openslides-autoupdate-service/openslides-autoupdate-service /
+
```
ostcar, line 80: I think this line can be removed. It is the same as line 77

ostcar (changes requested):
Only to typo-like issues

Dockerfile
```diff
@@ -11,6 +11,15 @@ ENV APP_CONTEXT=${CONTEXT}
...
 
+## Setup
+ARG CONTEXT
+
+WORKDIR /app/openslides-autoupdate-service
+ENV APP_CONTEXT=${CONTEXT}
+
+## Install
+RUN apk add git --no-cache
+
```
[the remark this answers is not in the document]
Janmtbehrens: Good catch, this must've come from a weird merge error. I removed it

Dockerfile
Janmtbehrens: Same as above

ostcar (approved):

Janmtbehrens:
> @Janmtbehrens is this still needed? Do you want to merge or close it?

It's still needed, but depends on all other Makefile Rework PRs to be reviewed and merged at the same time. Currently we are waiting for the Main Repository PR Review

PR #1258, opened 2025-09-12
Keep alive

ostcar:
This PR also contains some cleanup in the frist commit, that made the keep alive change easie

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/OpenSlides/openslides-autoupdate-service
 go 1.25.0
 
 require (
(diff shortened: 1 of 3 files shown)


ostcar:
@bastianjoel It just works with the client. So a change in the client is not necessary.

But the client creates log-message in the console for each heartbeat like `[autoupdate] from streams: 367969058 - organization_detail:subscription Array [ {}, {} ]`

You could change the log in the client, to skip empty messages. But this is independent from this change here.

bastianjoel:
> You could change the log in the client, to skip empty messages. But this is independent from this change here.

I will open a client issue for that.

PR #1261, opened 2025-09-14
Use snapshots

ostcar:
This uses snapshots to get all data at a specific version.

It needs this PR from openslides-go: https://github.com/OpenSlides/openslides-go/pull/89


ostcar:
@rrenkert could you make a performancetest of it? Maybe until 10 AM?

bastianjoel (approved):

PR #1280, opened 2025-10-21
Downgrade topic

rrenkert:


ostcar:
Can you still try to reproduce this? This update will go into main again, as soon, as I update all dependencies (what I do regularly)

ostcar:
You should try to reproduce the error. Then I can debug it (it is quite easy to debug a dependency with go.work)

PR #1363, opened 2026-03-07
Use go with multi postgres environment

ostcar:
@peb-adr This is a bit hard to test for me. Could you check this out and try to use it in a multi postgres environment?

If it works, we merge https://github.com/OpenSlides/openslides-go/pull/180. This will automatically generate PRs for vote and icc


peb-adr:
This is tested successfully and the changes in openslides-go are already included where needed.
This PR can in principle just be merged into main.
Only someone needs to decide which go pointer is correct for that

peb-adr:
@ostcar If this is immediately clear to you you may merge at will.
Otherwise I will sort it out - probably within the current week.

ostcar:
> Only someone needs to decide which go pointer is correct for that

I don't understand, what you mean with that

peb-adr:
> > Only someone needs to decide which go pointer is correct for that
> 
> I don't understand, what you mean with that

@ostcar Basically I meant to resolve the merge conflict by using the openslides-go version referenced on main.

Did that now, enabling automerge.

PR #1419, opened 2026-05-04
Add Target for prod Image with loacal openslides-go

LinKaKling:
This should make it possible to create prod images with a local openslides-go.
Adds prod-gowork
(Preparation for changes in Openslides/dev/docker/build.sh)


ostcar (approved):

peb-adr (approved):
As discussed, this seems to be the best possible solution.
LGTM

PR #1428, opened 2026-05-17
Remove required to see for polls

### LINKED ISSUE #1427: Don't show user names, if they have voted

ostcar:
@emanuelschuetze this is the change, we talked about. At the moment, a user can see the names of users, that have voted, even when the user does not have `user.can_see`.

I asked @normanjaeckel and he thinks, that it is not generally necessary to see the names of users, that have voted. There are cases, where this is necessary, but in this cases, its ok, to give the permission `user.can_see`.

@bastianjoel If I understand it correctly, the client already hided user-names in this cases.


ostcar:
I implemented this chance in this PR for the new vote serivce: https://github.com/OpenSlides/openslides-autoupdate-service/pull/1443

It will be a bit more difficult, to keep the current behavior in the new vote service. If you don't say anything about this, this change will be part of the new vote service.

PR #1451, opened 2026-07-13
Add `meeting_poll_default` collection

bastianjoel:


internal/restrict/collection/meeting_poll_default.go
```diff
@@ -0,0 +1,48 @@
...
+type MeetingPollDefault struct{}
+
+// Name returns the collection name.
+func (m MeetingPollDefault) Name() string {
+	return "meeting_poll_default"
+}
+
+// MeetingID returns the meetingID for the object.
+func (m MeetingPollDefault) MeetingID(ctx context.Context, ds *dsfetch.Fetch, id int) (int, bool, error) {
+	return id, true, nil
```
ostcar: This does not seem right. This returns `meeting_poll_default/id`, but `meeting_poll_default/meeting_id` is expected.

Here is an example:

https://github.com/OpenSlides/openslides-autoupdate-service/blob/e83b202e7b0ce51e7bc43600c0a830a8ec158035/internal/restrict/collection/chat_group.go#L28-L34

internal/restrict/collection/meeting_poll_default_test.go
```diff
@@ -0,0 +1,48 @@
...
+		false,
+		`---
+		meeting/30:
+			id: 1
+			committee_id: 300
+
+		meeting_poll_default/5:
+			meeting_id: 30
+		`,
+		withElementID(30),
```
ostcar: This is wrong. You either have to use `withElementID(5)` (the id of the meeting_poll_default-element), or you have to change the id to `1`(`meeting_poll_default/1`). `1` is the default. Then you do not need the `withElement` argument.

internal/restrict/collection/meeting_poll_default_test.go
```diff
@@ -0,0 +1,48 @@
...
+		group/7/meeting_user_ids: [10]
+		meeting_user/10:
+			user_id: 1
+			meeting_id: 30
+		user/1/meeting_user_ids: [10]
+
+		meeting_poll_default/5:
+			meeting_id: 30
+		`,
+		withElementID(30),
```
ostcar: Here the same

ostcar (changes requested):

ostcar (approved):

PR #1460, opened 2026-08-19
Update os-go

bastianjoel:


ostcar:
Why did that not happen automaticly?

Could you also call "go mod tidy" (not so important)

ostcar (approved):

bastianjoel:
`go generate` does not work in the Autoupdate currently

